PR #119, opened 2026-02-24
feat(scheduler,biometrics): clock validation, sleep records, and plugin module system

ng:
## Summary

- **System clock validation**: Pod can boot with RTC reset to ~2010; scheduler now polls until the system clock is valid (year ≥ 2024) before starting, preventing cron jobs from firing at wrong times
- **Reboot job fix**: `executeReboot()` is now async and rejects on failure, so scheduler error-handling surfaces `systemctl reboot` failures instead of silently swallowing them
- **Pre-prime reboot**: Auto-schedule a device reboot 1 hour before daily priming to ensure a clean state (matches free-sleep behaviour)
- **Power-based sleep records**: Detect ON→OFF power transitions inside an atomic `db.transaction()` to prevent TOCTOU double-creation; stub sleep records are written to `biometrics.db` (sessions < 5 min ignored)
- **Biometrics DB separation**: Move time-series data (`vitals`, `sleep_records`, `movement`) to a dedicated `biometrics.db` with its own Drizzle connection (WAL + 5s busy_timeout for concurrent module writes). Greenfield project — tables were dropped and recreated cleanly; no data migration needed.
- **Bundled Python modules**: Ship two production-ready sidecars — `piezo-processor` (HR/HRV/breathing from piezo FFT) and `sleep-detector` (session boundaries from capacitance presence) — with systemd service templates and installation automation
- **Squashed migrations**: Replaced 4 incremental migrations (create → alter → drop) with a single clean `0000` reflecting the current schema. No live deployments, so no migration path needed.
- **ADR consolidation**: Merged shallow one-decision ADRs into two coherent documents (core stack, developer tooling); added ADR 0012 for the biometrics plugin architecture

## Scope Breakdown

| Area | Summary |
|------|---------|
| `instrumentation.ts` | Add `waitForValidSystemDate()` polling before scheduler init |
| `src/scheduler/jobManager.ts` | Async `executeReboot()` with proper rejection; add `schedulePrimePreReboot()` |
| `src/hardware/deviceStateSync.ts` | Wrap SELECT+UPSERT in transaction; detect power transitions; write sleep records |
| `src/db/` | New `biometrics-schema.ts` + `biometrics.ts`; separate Drizzle connection; squashed to single migration |
| `modules/` | `piezo-processor` and `sleep-detector` Python sidecars with manifests and systemd units |
| `scripts/install` | Biometrics DB migration; Python venv setup with python3-venv guard; module installation |
| `docs/adr/` | Consolidate 5 ADRs into 2; add ADR 0012 for plugin/sidecar architecture |

## Design rationale

**Why a separate biometrics.db?**
Signal processing (HR from 500 Hz piezo data) is compute-heavy and better suited to Python/Rust than Node.js. Modules are independent OS processes — a crash in the HR module doesn't affect the core app. The SQLite schema is the public contract; any module that writes matching rows works automatically with the existing UI. Two DB files lets biometrics data be cleared, backed up, or handed off independently of config/state.

**Why squash migrations?**
No live deployments. The original history created biometrics tables in `sleepypod.db` and then dropped them two migrations later — noise with no benefit. One clean migration is easier to reason about.

**Why not WebSockets for module transport?**
Modules run on the same device and can read `/persistent/*.RAW` directly — no proxy needed today. WebSockets are not ruled out for future use (e.g. modules on a separate host, or real-time sensor push to the UI).

## Test plan

- [ ] Start with system clock set to 2015 — scheduler should poll and wait, not crash
- [ ] Start with valid clock — scheduler should start immediately
- [ ] Enable `primePodDaily` — verify `daily-prime` and `prime-prereboot` jobs appear in logs, prereboot 1hr before prime
- [ ] Enable `rebootDaily` — verify `daily-reboot` calls `systemctl reboot`
- [ ] Power pod on, wait 10s, power off — verify `sleep_records` row created in `biometrics.db`
- [ ] Power cycle < 5s — verify no sleep record created
- [ ] Run `pnpm db:generate && pnpm db:biometrics:generate` — verify single clean migrations for each DB
- [ ] Check module services: `systemctl status sleepypod-piezo-processor sleepypod-sleep-detector`

## Release Notes

* **New Features**
  * Sleep detection and session tracking with bed occupancy analysis
  * Biometric signals: heart rate, HRV, and breathing rate measurements
  * Movement tracking with per-bed-side granularity
  * Modular biometrics plugin system for extensibility
  * Power state tracking with automatic session boundaries
  * System clock validation at startup for reliability

* **Infrastructure**
  * Dedicated biometrics database for improved data organization
  * Enhanced scheduler with pre-prime reboot capability

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,363 @@
-# sleepypod
+# SleepyPod
+
+A self-hosted control system for [Eight Sleep](https://www.eightsleep.com/) Pod mattress covers (Pod 3, 4, and 5). Runs directly on the Pod's embedded Linux hardware, replacing the cloud dependency with a local-first web interface and scheduler.
+
+---
+
+## What it does
+
+- **Temperature scheduling** — set per-side temperature programs by day and time
+- **Power scheduling** — automatic on/off with optional warm-up temperature
+- **Alarm management** — vibration alarms with configurable intensity and pattern
+- **Biometrics** — heart rate, HRV, breathing rate, sleep session tracking, and movement from the Pod's own sensors
+- **Daily maintenance** — automated priming and system reboots on a schedule
+- **Local web UI** — accessible on your home network, no cloud required
+
+---
+
+## Architecture
+
+```mermaid
+graph TD
+    subgraph hw [Pod Hardware]
+        DAC["dac.sock<br/>Device Control"]
+        RAW["RAW Files<br/>/persistent/*.RAW"]
+    end
+
+    subgraph core [SleepyPod Core - Next.js]
+        UI[React UI]
+        API[tRPC API]
+        SCHED["Scheduler<br/>node-schedule"]
+        SYNC[DeviceStateSync]
+        DB1[("sleepypod.db<br/>Config & State")]
+    end
+
+    subgraph mods [Biometrics Modules]
+        PP["piezo-processor<br/>Python sidecar"]
+        SD["sleep-detector<br/>Python sidecar"]
+        DB2[("biometrics.db<br/>Vitals & Sleep")]
+    end
+
+    UI -->|queries| API
+    API -->|reads/writes| DB1
+    API -->|reads| DB2
+    SCHED -->|commands| DAC
+    SYNC -->|status events| DB1
+    RAW -->|tails CBOR| PP
+    RAW -->|tails CBOR| SD
+    PP -->|writes rows| DB2
+    SD -->|writes rows| DB2
+    DAC -.->|status updated| SYNC
+```
+
+### Biometrics data flow
+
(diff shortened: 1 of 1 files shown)


ng:
@coderabbitai All feedback has been addressed. Four comments are still showing as unresolved but are stale against earlier commits:

- **arrow-parens ()**: `executeReboot` was rewritten as `async/await` — the `.catch(error => {` chain no longer exists
- **data migration ()**: file was deleted — all migrations were squashed into a single clean `0000_zippy_mephistopheles.sql`
- **`system_health` UNIQUE constraint (ADR + `sleep-detector`)**: `.unique()` added to the Drizzle schema column, `CREATE UNIQUE INDEX system_health_component_unique` is present in the migration — `ON CONFLICT(component)` upserts in both modules will work correctly

Please re-review the latest commits.

PR #300, opened 2026-03-31
feat: 24/7 temperature mode with keepalive and stale display fix

### LINKED ISSUE #297: feat: 24/7 temperature mode — pod auto-offs after 8hr duration limit
## User Report

From Discord (Drizzy):

> It seems like the only way to get it to start cooling is to lower the temp for like 30 seconds, then put the temp back to the level you previously had it. It's like it doesn't stay constantly cooled to the temp I set it to. When I open the app it says the current temp is at 75°F and it's warming to my set temp, but then I lower the temp by 1°F and wait a sec and it'll say it's actually 82°F. Not sure if there is like a timer or something, but preferably I'd like to keep the cover 24/7 cooled to the temp I have it set to.

User workaround: setting a schedule to adjust temp ±1°F every 6 hours to keep it on.

## Root Cause

The firmware requires a `duration` parameter with every temperature command. After the duration expires, the pod returns to neutral (82.5°F / level 0). There is no "indefinite" mode.

**Current behavior:**
- `DEFAULT_HEATING_DURATION = 28800` (8 hours) — `src/hardware/types.ts:100`
- Every `setTemperature()` call sends the duration via `LEFT_TEMP_DURATION` / `RIGHT_TEMP_DURATION` hardware commands — `src/hardware/client.ts:190-196`
- After 8 hours, firmware autonomously returns to neutral — the app has no visibility into this event
- The DB still shows the old `targetTemperature` until the next `getStatus()` poll, so the UI appears to show "warming to 72°F" when the pod is actually at 82°F neutral

**Why the ±1°F trick works:** Adjusting temperature sends a fresh `setTemperature` command, which resets the firmware's 8-hour duration timer.

### 1. No 24/7 mode (feature gap)
There's no way for a user to keep the pod continuously cooled/heated. Max duration is 8 hours, then auto-off.

### 2. Stale temperature display after duration expires (bug)
When the firmware duration expires and returns to neutral, the app still shows the old target temperature because:
- The firmware doesn't notify the app when duration expires
- DB `deviceState.targetTemperature` is only updated on user actions or status polls
- `isPowered` is derived from `targetLevel !== 0` in `getStatus()`, so it only corrects on the next hardware poll

### 24/7 mode — software keepalive (most feasible)
Since the firmware always requires a finite duration, implement a software keepalive:
- Add an "Always On" toggle per side (settings or TempScreen)
- When enabled, a background job periodically re-sends the current target temperature (e.g., every 6 hours) to reset the firmware duration timer
- Could reuse the scheduler/jobManager infrastructure
- The `autoOffEnabled` / `autoOffMinutes` fields already exist in `side_settings` schema — repurpose or add a complementary `alwaysOn` field

### Stale display fix
- Poll `getStatus()` on a timer (already exists in DAC monitor) and compare firmware `heatingDuration` (seconds remaining) against a threshold
- When `heatingDuration` reaches 0 and `targetLevel` is 0 but DB says `isPowered: true`, auto-correct the DB state
- Or: show the firmware's `heatingDuration` countdown in the UI so users know when auto-off will occur (the `EnvironmentInfoPanel` already has `secondsRemaining` prop support)

## Key Files

| File | Relevance |
|------|-----------|
| `src/hardware/types.ts:100` | `DEFAULT_HEATING_DURATION = 28800` |
| `src/hardware/client.ts:167-197` | `setTemperature()` always sends duration |
| `src/server/routers/device.ts:181-246` | `setTemperature` mutation (debounced) |
| `src/hardware/responseParser.ts:39,46` | Parses `heatingDuration` from firmware |
| `src/db/schema.ts:41-42` | `autoOffEnabled` / `autoOffMinutes` fields |
| `src/components/EnvironmentInfo/EnvironmentInfoPanel.tsx` | Already supports `secondsRemaining` display |
| `src/scheduler/jobManager.ts` | Could host the keepalive job |

ng:
## Summary
- Adds `alwaysOn` field to side settings with a keepalive service that re-sends temperature every 6 hours to prevent firmware's 8-hour auto-off
- Fixes stale temperature display when firmware duration expires — auto-corrects DB state when `targetLevel=0` and `heatingDuration=0`
- New `setAlwaysOn` mutation and startup/shutdown wiring

Closes #297

## Test plan
- [ ] Enable always-on for a side — verify keepalive timer starts
- [ ] Disable always-on — verify timer stops
- [ ] Verify stale display fix: when firmware returns to neutral, DB updates to `isPowered: false`
- [ ] Run migration on fresh DB

## Summary by CodeRabbit

* **New Features**
  * "Always On" temperature mode: devices stay powered and have target temperatures re-sent periodically to maintain settings.
  * New API to enable/disable "Always On" per side.

* **Chores**
  * Database schema updated to persist the "Always On" setting.

* **Stability**
  * Keepalive lifecycle now starts/shuts reliably during startup and shutdown.

### DIFF
--- a/instrumentation.ts
+++ b/instrumentation.ts
@@ -20,6 +20,7 @@ import { closeDatabase, closeBiometricsDatabase } from '@/src/db'
 import { getDacMonitor, shutdownDacMonitor } from '@/src/hardware/dacMonitor.instance'
 import { startPiezoStreamServer, shutdownPiezoStreamServer } from '@/src/streaming/piezoStream'
 import { startBonjourAnnouncement, stopBonjourAnnouncement } from '@/src/streaming/bonjourAnnounce'
+import { initializeKeepalives, shutdownKeepalives } from '@/src/services/temperatureKeepalive'
 
 let isInitialized = false
 let isShuttingDown = false
@@ -41,6 +42,14 @@ async function gracefulShutdown(signal: string): Promise<void> {
     process.exit(1)
   }, 10_000)
   forceExitTimer.unref()
+
+  // Step 0: Stop keepalive timers
+  try {
+    shutdownKeepalives()
+  }
+  catch (error) {
+    console.error('Error shutting down keepalives:', error)
+  }
 
   // Step 1: Shutdown scheduler (waits for in-flight jobs internally)
   try {
@@ -251,6 +260,9 @@ export async function initializeScheduler(): Promise<void> {
 
     // Start DAC monitor (non-blocking — waits for frankenfirmware to connect)
     initializeDacMonitor()
+
+    // Initialize temperature keepalive timers for sides with alwaysOn enabled
+    initializeKeepalives()
 
     // Start piezo WebSocket stream server (non-blocking)
     try {
--- a/src/db/migrations/0004_shallow_tomorrow_man.sql
+++ b/src/db/migrations/0004_shallow_tomorrow_man.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `side_settings` ADD `always_on` integer DEFAULT false NOT NULL;
--- a/src/db/migrations/meta/0004_snapshot.json
+++ b/src/db/migrations/meta/0004_snapshot.json
@@ -1,0 +1,781 @@
+{
+  "version": "6",
+  "dialect": "sqlite",
+  "id": "48a5ab38-529c-496f-8052-b8760da45bf8",
+  "prevId": "a578bda9-6bb5-48fe-be74-65002e304ccd",
+  "tables": {
+    "alarm_schedules": {
+      "name": "alarm_schedules",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "integer",
+          "primaryKey": true,
+          "notNull": true,
+          "autoincrement": true
+        },
+        "side": {
+          "name": "side",
+          "type": "text",
+          "primaryKey": false,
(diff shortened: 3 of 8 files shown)


ng (commented):
## Adversarial Code Review — Full Depth (Sonnet + Opus)

### Fixed (2)
- **initializeKeepalives skip** — Now starts timer unconditionally for `alwaysOn=true` sides (was skipping powered-off sides, defeating the feature after server restart)
- **setInterval gap** — `startKeepalive` now fires immediately on enable (was waiting 6h for first tick)

### Disputed (2, author decision needed)
- **`durationExpired` naming** — Sonnet says misleading; Opus says accurate. Low risk either way.
- **`alwaysOn`/`autoOff` conflict** — No runtime autoOff yet. Address mutual exclusion when both features coexist.

### Deferred (4)
- No test coverage for keepalive service (Minor)
- Keepalive can't recover after firmware timeout — pod stays off permanently despite `alwaysOn=true` (Major, needs design)
- `setAlwaysOn` endpoint bypasses future constraints added to `updateSide` (Major)
- Duplicate `setAlwaysOn` endpoint — keep but add transaction wrapper (Minor)

### Rejected (2)
- Stale display on firmware transients — false positive (both models confirm firmware doesn't produce this state)
- `client.connect()` redundant — factually incorrect (every call site uses this pattern)

**Recommendation**: Approve with deferred items noted

PR #301, opened 2026-03-31
feat: auto-off when no bed presence detected

### LINKED ISSUE #257: feat: auto-off when no bed presence detected
## Summary

Automatically power off a side when no presence is detected for a configurable duration. Saves energy and prevents the pod from heating/cooling an empty bed indefinitely.

## Behavior

When the pod detects no occupancy on a side for longer than the configured timeout, it powers off that side. When the user gets back in bed, they would need to manually power on or let the next scheduled event handle it.

### Detection source

Use the existing capacitance-based presence detection from `SensorStreamService` / `sleep-detector`. The piezoStream WS already broadcasts `capSense2` frames, and `sleep-detector` tracks session boundaries (bed entry/exit). Either source could trigger the auto-off:

- **Option A**: Server-side — `sleep-detector` writes a sleep record on bed exit. A new watcher in sleepypod-core monitors for exit events and starts a countdown timer.
- **Option B**: Server-side — DacMonitor or a new service polls the latest capSense data and tracks continuous absence.

Option A is simpler since sleep-detector already handles the presence logic.

## Configuration

New fields in `side_settings`:

| Field | Type | Default | Description |
|-------|------|---------|-------------|
| `autoOffEnabled` | boolean | false | Enable auto-off for this side |
| `autoOffMinutes` | integer | 30 | Minutes of no presence before powering off (range: 5-120) |

### Settings UI

Add to the per-side settings panel (iOS `SideSettingsForm`, web `SideSettings`):
- Toggle: "Auto-off when empty"
- Duration picker: 5 / 10 / 15 / 30 / 45 / 60 / 90 / 120 minutes

## Implementation sketch

1. **Schema**: Add `auto_off_enabled` and `auto_off_minutes` to `side_settings` table
2. **Watcher service**: New module that listens for bed-exit events (from sleep-detector's `sleep_records` table or from live capSense frames) and starts a per-side countdown timer
3. **Timer fires**: Call `device.setPower(side, false)` + `broadcastMutationStatus`
4. **Cancellation**: Timer resets if presence is re-detected before it expires
5. **Interaction with run-once**: If a run-once session is active, auto-off should probably be suppressed (the user explicitly chose to run a curve)
6. **Interaction with schedules**: A scheduled power-on should cancel any pending auto-off timer

## Edge cases

- User gets up briefly (bathroom) → timer starts → comes back within timeout → timer cancelled, no power-off
- Both sides empty → each side has its own independent timer
- Pod is already off → no-op (don't start timer for an already-off side)
- Auto-off fires during a run-once session → should it cancel the session too, or just suppress auto-off?

## Applies to

- Core: `src/db/schema.ts`, `src/server/routers/settings.ts`, new watcher service
- iOS: `SideSettingsForm.swift`
- Web: side settings component

ng:
## Summary
- New auto-off watcher service that monitors sleep_records for bed-exit events and powers off after configurable timeout
- Per-side countdown timers with reset on re-entry, suppression during run-once sessions
- Web UI: toggle + duration picker (5–120 min) in side settings form
- Wired into server startup/shutdown lifecycle

Closes #257

## Test plan
- [ ] Enable auto-off with 5 min timeout — verify side powers off after no presence
- [ ] Verify timer resets on bed re-entry
- [ ] Verify auto-off suppressed during active run-once session
- [ ] Verify duration picker UI renders correctly

## Summary by CodeRabbit

* **New Features**
  * Per-side automatic power-off: enable/disable auto-off per side with configurable timeout in Settings (toggle + duration picker).
  * Settings apply immediately: changes to auto-off or timeout take effect without needing a restart.
  * Auto-off respects active use: timers pause/cancel when a side is powered on or detects return to bed, reducing unwanted shutdowns.

### DIFF
--- a/instrumentation.ts
+++ b/instrumentation.ts
@@ -76,9 +76,9 @@ async function gracefulShutdown(signal: string): Promise<void> {
     console.error('Error stopping Bonjour:', error)
   }
 
-  // Step 4: Stop auto-off watcher
+  // Step 4: Stop auto-off watcher (await in-flight power-off calls)
   try {
-    stopAutoOffWatcher()
+    await stopAutoOffWatcher()
   }
   catch (error) {
     console.error('Error stopping auto-off watcher:', error)
--- a/instrumentation.ts
+++ b/instrumentation.ts
@@ -21,6 +21,7 @@ import { getDacMonitor, shutdownDacMonitor } from '@/src/hardware/dacMonitor.ins
 import { startPiezoStreamServer, shutdownPiezoStreamServer } from '@/src/streaming/piezoStream'
 import { startBonjourAnnouncement, stopBonjourAnnouncement } from '@/src/streaming/bonjourAnnounce'
 import { initializeKeepalives, shutdownKeepalives } from '@/src/services/temperatureKeepalive'
+import { startAutoOffWatcher, stopAutoOffWatcher } from '@/src/services/autoOffWatcher'
 
 let isInitialized = false
 let isShuttingDown = false
@@ -75,7 +76,15 @@ async function gracefulShutdown(signal: string): Promise<void> {
     console.error('Error stopping Bonjour:', error)
   }
 
-  // Step 4: Shutdown DAC monitor
+  // Step 4: Stop auto-off watcher
+  try {
+    stopAutoOffWatcher()
+  }
+  catch (error) {
+    console.error('Error stopping auto-off watcher:', error)
+  }
+
+  // Step 5: Shutdown DAC monitor
   try {
     await shutdownDacMonitor()
   }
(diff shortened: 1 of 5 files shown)


ng (commented):
## Adversarial Code Review — Standard Depth

### Fixed (3)
- **Missing broadcast** — `powerOffSide` now calls `broadcastMutationStatus(side, { isPowered: false, targetLevel: 0 })` so clients update immediately
- **Timer fires on re-entry** — Callback re-fetches latest sleep record and verifies `leftBedAt` matches before firing, preventing auto-off while user is back in bed
- **DB query reduction** — Hoisted `getAutoOffConfig()` into `poll()` to halve config reads per cycle

### Deferred (4)
- Scheduled power-on should cancel auto-off timer (Major — issue #257 requirement gap)
- `stopAutoOffWatcher` can't cancel in-flight `powerOffSide` calls during shutdown (Major)
- Non-standard `autoOffMinutes` values have no UI representation (Minor)
- Missing index on `(side, leftBedAt)` for `getLatestSleepRecord` (Nit)

### Rejected (1)
- Timer state reset race — mathematically impossible (time monotonically increases for same `leftBedAt`)

**Recommendation**: Approve with deferred items noted

PR #308, opened 2026-04-01
feat: auto-unblock internet during update check

### LINKED ISSUE #306: feat: auto-unblock internet during update check
## Context

The pod blocks WAN internet by default via iptables for privacy. Users can toggle internet access via the InternetToggleCard in the web UI or `system.setInternetAccess` API. However, when a user presses "Check for updates," the update check silently fails if internet is blocked — no error, no feedback, nothing happens.

A user reported this exact scenario: pressed update, confirmed, nothing happened. Turned out internet was blocked and they had to manually unblock via an external script (didn't know about the built-in toggle).

## Proposed behavior

When the user initiates an update check:

1. **Check internet status** first via `system.internetStatus`
2. If blocked, **temporarily unblock** internet for the duration of the update check/download
3. After the update completes (or fails), **re-block** internet automatically
4. If the unblock fails, show a clear error: "Internet access is blocked. Enable internet access to check for updates."

This is implied behavior — pressing "check for updates" inherently requires internet. The user shouldn't have to know about iptables.

## Alternative (simpler)

If auto-unblock feels too magical:
- Just check internet status before the update request
- If blocked, show a prompt: "Internet is currently blocked. Temporarily allow internet to check for updates?" with Allow/Cancel buttons
- On Allow, unblock → check → re-block

## Key files
- `src/components/` — find the update check UI component
- `src/server/routers/system.ts` — `internetStatus` and `setInternetAccess` endpoints
- `src/server/routers/app.ts` — likely has the update check logic
- `src/hardware/iptablesCheck.ts` — iptables validation

ng:
## Summary
- When user initiates update check with internet blocked, shows prompt to temporarily allow internet
- On allow: unblocks internet, proceeds with update, re-blocks automatically after completion/failure
- Uses `system.internetStatus` and `system.setInternetAccess` APIs
- Internet re-blocked in both success and failure paths via ref tracking

Closes #306

## Test plan
- [ ] With internet blocked: press update → verify prompt appears
- [ ] Press "Allow & Update" → verify internet unblocks, update proceeds, internet re-blocks
- [ ] Press Cancel → verify no changes
- [ ] With internet already allowed: press update → verify normal flow (no prompt)

## New Features
* Update process now detects when internet access is blocked before proceeding
* Displays a prompt asking users to allow internet access when needed
* Added "Allow & Update" button for users to grant temporary internet access
* Automatically re-restricts internet access after the update completes

### DIFF
--- a/src/components/status/UpdateCard.tsx
+++ b/src/components/status/UpdateCard.tsx
@@ -2,7 +2,7 @@
 
 import { useEffect, useRef, useState } from 'react'
 import { trpc } from '@/src/utils/trpc'
-import { CheckCircle, Download, Loader2, RefreshCw, AlertTriangle } from 'lucide-react'
+import { CheckCircle, Download, Loader2, RefreshCw, AlertTriangle, Globe } from 'lucide-react'
 
 /**
  * UpdateCard — shows current version and provides a trigger to update the pod software.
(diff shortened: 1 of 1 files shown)


ng (commented):
## Adversarial Code Review — Standard Depth

### Fixed (1)
- **Unmount during polling leaves WAN open** — Added `reblockIfNeeded` to useEffect cleanup. If user navigates away after "Allow & Update", internet is now re-blocked on unmount.

### Rejected (3 — false positives)
- Cancel leak — Cancel button is removed from DOM before unblock resolves; path unreachable
- startUpdate throws past catch — startUpdate never propagates; its own catch always fires
- Stale cache (10s) — staleTime defaults to 0 in this project; fresh fetch always issued

### Noted (3 — architectural, not this PR)
- `blockWan()` in system.ts omits mDNS rules vs install script (pre-existing Major)
- sp-update double iptables management — design choice, UI manages WAN lifecycle
- Silent re-block failure — best-effort is acceptable for this flow

**Recommendation**: Approve

PR #337, opened 2026-04-03
fix: persist free-sleep/sleepypod switch across reboots

### LINKED ISSUE #334: back to free sleep script broken
after using the sp-freesleep command it does switch back but once the pod restarts it just reverts to sleepypod

ng:
## Summary
- `sp-freesleep` now disables `sleepypod.service` and enables `free-sleep.service` / `free-sleep-stream.service` so the choice survives reboots
- `sp-sleepypod` does the reverse — disables free-sleep services and re-enables sleepypod

Previously both commands only stopped/started services at runtime without toggling `systemctl enable/disable`, so every reboot reverted to sleepypod regardless of which mode the user selected.

Fixes #334

## Test plan
- Run `sp-freesleep`, verify free-sleep is active
- Run `systemctl is-enabled sleepypod.service` → should be `disabled`
- Run `systemctl is-enabled free-sleep.service` → should be `enabled`
- Reboot pod → free-sleep should still be running
- Run `sp-sleepypod`, verify sleepypod is active and enabled states are reversed
- Reboot pod → sleepypod should still be running

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -939,6 +939,7 @@ echo "Switching to free-sleep..."
 systemctl stop sleepypod.service 2>/dev/null || true
 for svc in sleepypod-piezo-processor sleepypod-sleep-detector sleepypod-environment-monitor sleepypod-calibrator; do
   systemctl stop "$svc.service" 2>/dev/null || true
+  systemctl disable "$svc.service" 2>/dev/null || true
 done
 # Persist across reboots
 systemctl disable sleepypod.service 2>/dev/null || true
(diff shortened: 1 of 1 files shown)


ng (commented):
Clean change, logic is correct for the main services. One gap:

`sp-freesleep` stops the 4 biometrics modules at runtime but doesn't `disable` them. On reboot they'll auto-start alongside free-sleep since they're still enabled. Same in reverse — `sp-sleepypod` should re-enable them.

```diff
# sp-freesleep — add disable to the existing stop loop:
 for svc in sleepypod-piezo-processor sleepypod-sleep-detector sleepypod-environment-monitor sleepypod-calibrator; do
   systemctl stop "$svc.service" 2>/dev/null || true
+  systemctl disable "$svc.service" 2>/dev/null || true
 done

# sp-sleepypod — add enable to the existing restart loop:
 for svc in sleepypod-piezo-processor sleepypod-sleep-detector sleepypod-environment-monitor sleepypod-calibrator; do
   systemctl restart "$svc.service" 2>/dev/null || true
+  systemctl enable "$svc.service" 2>/dev/null || true
 done
```

Otherwise LGTM.

PR #479, opened 2026-04-25
fix: streaming buffer bounds + WebSocket cleanup (#324)

### LINKED ISSUE #324: Streaming: WebSocket resource management & backpressure
## Summary

Major resource management issues in the WebSocket streaming layer that can cause memory exhaustion and event loop blocking on the embedded device.

### 🟡 `frameIndex` grows without bound
**File**: `src/streaming/piezoStream.ts:602`

Every decoded frame pushes to `frameIndex` (16 bytes/entry), only cleared on RAW file change. 24h at 50fps = 4.3M entries = ~70MB on a constrained device.

**Fix**: Cap to last N minutes of entries (seek only supports 30s window via `SEEK_MAX_DURATION_S`), or use a ring buffer.

### 🟡 `findLatestRaw` does synchronous I/O every 10ms
**File**: `src/streaming/piezoStream.ts:192-207`

`readdirSync` + `statSync` per file on every tick when clients are connected. Blocks the event loop 100 times/second.

**Fix**: Cache the latest RAW path, re-scan on a longer interval (5s), or use `fs.watch` on the directory.

### 🟡 No backpressure handling on WebSocket sends
**File**: `src/streaming/piezoStream.ts:618-623`

Frames sent without checking `ws.bufferedAmount`. A slow/stalled client causes unbounded server-side memory growth.

**Fix**: Check `client.bufferedAmount` before sending; skip if above threshold (e.g., 1MB).

### 🟡 No `maxPayload` on WebSocket server
**File**: `src/streaming/piezoStream.ts:516`

No `maxPayload` limit — a malicious client could send oversized messages to exhaust memory. Also no rate limiting on seek requests (each reads up to 64MB synchronously).

**Fix**: Add `{ maxPayload: 1024 }` (client messages are tiny JSON). Rate-limit seek requests per-client.

### 🟢 Seek reads up to 64MB synchronously into memory
**File**: `src/streaming/piezoStream.ts:421-431`

`Buffer.alloc(64MB)` + `fs.readSync` blocks the event loop during client seek requests, stalling all other clients.

**Fix**: Read in smaller chunks with `setImmediate` between, or stream since seek is capped at 30s.

---
*Filed by codebase audit (2026-04-03)*

ng:
## Summary

Closes #324. The WebSocket streaming layer had three resource-management holes that
could exhaust the pod's ~1.3 GB of free memory on long-running streams.

### Per-finding fixes (all in `src/streaming/piezoStream.ts`)

- **`frameIndex` unbounded growth** (was line 602) — pushed one entry per decoded
  frame, only cleared when the RAW file rotated. 24 h at 50 fps = ~4.3 M entries /
  ~70 MB. Introduced `appendFrameIndex()` which evicts anything older than
  `SEEK_MAX_DURATION_S + 10s` on every append. Seek is already capped at 30 s via
  `SEEK_MAX_DURATION_S`, so older entries are unreachable anyway. The eviction is
  amortized O(1) — a single leading `splice`, not a per-entry `shift`.

- **No backpressure on outbound sends** (was lines 618-623) — raw `client.send()`
  without inspecting `bufferedAmount` meant a single stalled client could pin
  unbounded server memory. Added `sendWithBackpressure()` which drops the frame
  (and bumps a per-client counter) if `bufferedAmount > 1 MB`. All three
  send-sites (live broadcast, seek replay, `broadcastFrame`) now route through
  it. The close handler logs the drop count on disconnect so stalled clients are
  visible in the journal.

- **Per-client cleanup on close** — `clientSubscriptions` was a `WeakMap` so
  cleanup waited for a GC cycle. Replaced with a regular `Map`, added a
  `clientDroppedFrames` `Map` for the backpressure counter, and a
  `cleanupClient()` helper the `close` handler calls explicitly. State is freed
  deterministically when the socket closes.

- **No `maxPayload` on WebSocketServer** (bonus, issue's 4th finding) — added
  `maxPayload: 1024`. All client→server messages (`subscribe`, `get_time_range`,
  `seek`) are tiny JSON and the largest is ~200 bytes, so this is well above
  normal traffic but low enough that a malicious client can't allocate large
  buffers before the framing layer closes the connection.

### Memory profile notes

- Before: frameIndex at steady state with no file rotation was unbounded. 24 h
  streaming session accumulated ~70 MB of index entries plus whatever a stalled
  client had buffered (also unbounded).
- After: frameIndex caps at roughly `(peak fps) x FRAME_INDEX_RETENTION_S` entries
  (50 × 40 ≈ 2 000 entries ≈ 32 KB at 16 B/entry). Per-client buffered-send is
  hard-capped at 1 MB before we start dropping frames, so N connected clients are
  bounded at roughly N MB of send-buffer memory.
- Constants sit next to each other at the top of the file (`FRAME_INDEX_RETENTION_S`,
  `MAX_BUFFERED_BYTES`, `WS_MAX_PAYLOAD_BYTES`) so they are easy to tune.

### Deliberate non-goals

Issue #324 also lists two 🟡/🟢 findings not in the task description: (a) the 10 ms
`findLatestRaw` sync I/O loop and (b) the 64 MB synchronous seek read. Per the
"no drive-by refactors" constraint I left those as-is; they can be filed as
follow-up issues if still priority after this lands.

## Test plan

All covered by automated tests in `src/streaming/tests/piezoStream.test.ts`
(17 new tests, 60 tests affected by `vitest related`):

- [x] Bounded frame index retains entries inside the window and evicts older ones.
- [x] Frame index stays bounded over a synthetic 10x-window run (regression for
      the 24 h / 4.3 M-entry case).
- [x] Timestamps stay monotonic after eviction.
- [x] `cleanupClient` deletes both per-client maps; no-op safe on unknown clients.
- [x] `sendWithBackpressure` sends below threshold, drops above, resumes when the
      buffer drains, handles closed sockets and send-throws without crashing.
- [x] Boundary check: exactly at `MAX_BUFFERED_BYTES` is allowed (strict >).
- [x] Transport config sanity checks on `WS_MAX_PAYLOAD_BYTES` and
      `MAX_BUFFERED_BYTES` so future tuning stays in safe ranges.

Quality gates (all green):

- [x] `pnpm vitest` — 636 tests pass, 2 skipped.
- [x] `pnpm tsc` — no errors.
- [x] `pnpm lint` — no new warnings (one pre-existing warning in `stryker.config.mjs`).

### Manual verification (recommended on pod)

- [ ] Connect one WebSocket client, leave running for > `SEEK_MAX_DURATION_S + 10 s`,
      verify RSS stays flat (no index growth).
- [ ] Hold a client (simulate slow link), check `[sensorStream] Client disconnected
      (dropped N frames due to backpressure)` in the journal.
- [ ] Send a 2 KB client message, verify the server closes the connection.

## Summary by CodeRabbit

* **Bug Fixes**
  * WebSocket streaming now includes backpressure handling to prevent client buffer overload with dropped frame tracking.

* **Performance**
  * Frame index memory is automatically managed with configurable retention to prevent unbounded growth.
  * Improved WebSocket connection cleanup and client state management.
  * Optimized WebSocket payload configuration.

* **Tests**
  * Comprehensive test coverage added for streaming backpressure, frame eviction, and connection handling.

### DIFF
--- a/src/streaming/piezoStream.ts
+++ b/src/streaming/piezoStream.ts
@@ -340,7 +340,10 @@ function cleanupClient(ws: WebSocket): void {
  */
 function sendWithBackpressure(client: WebSocket, payload: string): boolean {
   if (client.readyState !== WebSocket.OPEN) return false
-  if (client.bufferedAmount > MAX_BUFFERED_BYTES) {
+  // Account for the pending payload — checking bufferedAmount alone lets each
+  // send push the buffer past MAX_BUFFERED_BYTES before the next call notices.
+  const payloadByteSize = Buffer.byteLength(payload)
+  if (client.bufferedAmount + payloadByteSize > MAX_BUFFERED_BYTES) {
     clientDroppedFrames.set(client, (clientDroppedFrames.get(client) ?? 0) + 1)
     return false
   }
@@ -495,6 +498,7 @@ function handleSeek(ws: WebSocket, targetTs: number): void {
     let bufPos = 0
     const maxTs = targetTs + SEEK_MAX_DURATION_S
     let done = false
+    let droppedDuringReplay = 0
 
     while (bufPos < seekBuffer.length && !done) {
       try {
@@ -520,13 +524,24 @@ function handleSeek(ws: WebSocket, targetTs: number): void {
             done = true
             break
           }
-          sendWithBackpressure(ws, JSON.stringify(frame))
+          if (!sendWithBackpressure(ws, JSON.stringify(frame))) {
+            droppedDuringReplay += 1
+          }
         }
       }
       catch (e) {
         if (e instanceof RangeError) break // incomplete record
         bufPos += 1 // skip malformed byte, try to resync
       }
+    }
+
+    if (ws.readyState === WebSocket.OPEN) {
+      // Report partial replays so the client can re-seek if needed instead of
+      // assuming it received the full window.
+      ws.send(JSON.stringify({
+        type: 'seek_complete',
+        ...(droppedDuringReplay > 0 && { incomplete: true, droppedFrames: droppedDuringReplay }),
+      }))
     }
   }
   catch {
@@ -536,10 +551,6 @@ function handleSeek(ws: WebSocket, targetTs: number): void {
       }
       catch { /* ignore */ }
     }
-  }
-
-  if (ws.readyState === WebSocket.OPEN) {
-    ws.send(JSON.stringify({ type: 'seek_complete' }))
   }
 }
 
@@ -792,6 +803,10 @@ export async function shutdownPiezoStreamServer(): Promise<void> {
     wss = null
     return new Promise<void>((resolve) => {
       server.close(() => {
+        // Drop per-client state explicitly — relying on per-socket close
+        // events leaks if any handler failed to fire.
+        clientSubscriptions.clear()
+        clientDroppedFrames.clear()
         console.log('[sensorStream] WebSocket server closed')
         resolve()
       })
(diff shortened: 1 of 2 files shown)


ng:
@coderabbitai review

ng:
@coderabbitai review

ng:
@coderabbitai full review

ng:
@coderabbitai full review

ng:
@coderabbitai full review

PR #482, opened 2026-04-25
fix: install script integrity + systemd hardening (#319)

### LINKED ISSUE #319: Security: Install script integrity verification & systemd hardening
## Summary

Critical and major security findings from codebase audit affecting the install/deploy pipeline and runtime service isolation.

### 🔴 Node.js tarball downloaded without integrity check
**File**: `scripts/install:320`

The Node.js binary is fetched with `curl -fSL "$NODE_URL" | tar -xz` — no SHA-256 verification. A compromised CDN or MITM installs a backdoored binary silently.

**Fix**: Fetch `SHASUMS256.txt` from `https://nodejs.org/dist/v${NODE_FULL}/`, verify hash with `sha256sum` before extraction.

### 🟡 Systemd service runs as `User=root` with no capability dropping
**File**: `scripts/install:522`

The Node.js process runs as UID 0. Only `dac.sock` access and iptables need root. A web app compromise gives full root shell.

**Fix**: Create a `sleepypod` system user, use `AmbientCapabilities=CAP_NET_ADMIN` for iptables, run iptables check as `ExecStartPre` before dropping privileges.

### 🟡 Production browser source maps expose TypeScript source on LAN
**File**: `next.config.mjs:4`

`productionBrowserSourceMaps: true` serves `.map` files publicly, revealing full source including API routes and internal paths.

**Fix**: Set to `false` or use server-side-only source maps with a private error monitoring service.

### 🟡 `iptables-save` failure silently swallowed in `unblock_wan`
**File**: `scripts/install:73-90`

`|| true` on `iptables-save` means if state can't be saved, `restore_wan` falls through to a minimal ruleset. Silent data loss.

**Fix**: Remove `|| true`, guard `[ -n "$SAVED_IPTABLES" ]` before proceeding.

### 🟡 `sp-update` saves only one generation of database backup
**File**: `scripts/install:822-824`

Fixed name `sleepypod.db.bak` — two rapid updates overwrites the first backup.

**Fix**: Use timestamped naming: `sleepypod.db.bak.$(date +%s)`.

---
*Filed by codebase audit (2026-04-03)*

ng:
Closes #319.

## Summary

Addresses the five security findings from the install/deploy pipeline audit.

### Per-finding fixes

- **Node.js tarball SHA-256 verification** (`scripts/install`) — downloads `SHASUMS256.txt` from the same `nodejs.org/dist/v${NODE_FULL}/` directory and verifies the tarball with `sha256sum --check --ignore-missing` (falls back to `shasum -a 256` where coreutils is absent) before extraction. Fails loud if hash mismatch, missing hash tools, or download failure.

- **Systemd service no longer runs as `User=root`** (`scripts/install`) — creates a dedicated `sleepypod` system user (member of `sleepypod` primary group, `dac` supplementary group for `dac.sock` access). The unit:
  - `User=sleepypod Group=sleepypod SupplementaryGroups=dac`
  - `AmbientCapabilities=CAP_NET_ADMIN CAP_NET_RAW` — only what the runtime needs for `checkAndRepairIptables()` and internet-control routes
  - `CapabilityBoundingSet` locked to those two caps
  - `ExecStartPre=+...` prefix keeps root for the `dac.sock` symlink + `sp-maintenance` (journal vacuum / pnpm store prune)
  - Adds `ProtectKernelLogs`, `ProtectControlGroups`, `ProtectClock`, `ProtectHostname`, `PrivateTmp`, `PrivateDevices`, `RestrictNamespaces`, `RestrictRealtime`, `RestrictSUIDSGID`, `LockPersonality`, `RestrictAddressFamilies`, `SystemCallArchitectures=native`
  - `ReadWritePaths` expanded to include `/etc/iptables` (runtime persists rules there)
  - Notable omissions: `ProtectHome` (would hide `/home/dac/sleepypod-core`), `MemoryDenyWriteExecute` (breaks V8 JIT)

- **`productionBrowserSourceMaps` disabled** (`next.config.mjs`) — set to `false`. Use a private error monitor + server-side source maps for production debugging instead.

- **`iptables-save` failure no longer silently swallowed** (`scripts/install`, `scripts/lib/iptables-helpers`, `scripts/bin/sp-update` inline fallback) — captures stderr and refuses to unblock WAN if `iptables-save` fails or returns empty output (previous behavior was to fall through to a minimal `block_wan` ruleset on restore → silent data loss).

- **`sp-update` DB backup now timestamped** (`scripts/bin/sp-update`) — uses `sleepypod.db.bak.$(date +%s)` matching the install script + `sp-maintenance` pruning pattern (keeps 3 most recent). Two rapid updates no longer clobber the previous backup.

### Follow-on changes

- `sp-uninstall` removes the `sleepypod` user (group preserved for `--keep-data` file ownership).
- `.env` owner becomes `root:sleepypod` with mode `640` so the unprivileged service user can read it.
- `/home/dac` is made traversable (`o+x`) so the sleepypod service can reach `/home/dac/sleepypod-core`.
- `/etc/iptables` pre-created so namespace setup doesn't fail on missing directory.

### Assumptions

- `sp-update` does not rewrite the systemd unit file (existing behaviour). Users on existing installs need a re-run of the full installer to pick up the hardened unit. This matches how other install-time config (avahi service file, journald conf) is managed.
- The stale `scripts/install:822-824` file:line reference in the issue actually points at `scripts/bin/sp-update:129-131` — the install script already timestamps its backup. Fix applied to the real location.

## Test plan

- [ ] Fresh install on Pod 4 (`curl -fsSL ... | sudo bash`): Node.js tarball + SHASUMS256.txt download, sha256 verify passes, extraction succeeds, `id sleepypod` returns UID < 1000 with home `/nonexistent`.
- [ ] `systemctl show sleepypod.service -p User,AmbientCapabilities,CapabilityBoundingSet,ReadWritePaths` shows `sleepypod`, `cap_net_admin cap_net_raw`, and includes `/etc/iptables`.
- [ ] `systemctl status sleepypod.service` running healthy; endpoints respond at `http://<pod>:3000/`.
- [ ] `curl http://<pod>:3000/somepage.js.map` returns 404 (source maps gated).
- [ ] Simulated MITM on Node.js tarball (swap for corrupted file) — install aborts with `Node.js tarball SHA-256 verification failed`.
- [ ] `iptables-save` failure simulated (`chmod -x /sbin/iptables-save` then run `sp-update`) — script aborts with `iptables-save failed — refusing to unblock WAN`, WAN stays blocked.
- [ ] Two `sp-update` runs in a row — both create distinct `sleepypod.db.bak.<ts>` files; `sp-maintenance` prunes to 3 eventually.
- [ ] Runtime `iptablesCheck.ts` repair path still works (kill an mDNS rule then hit `/api/health.system` — auto-repair succeeds despite dropped root).

## Release Notes

* **Security Improvements**
  * Service now runs as a dedicated unprivileged user with minimal capabilities instead of root.
  * Node.js installation now verifies checksums before extraction.

* **Bug Fixes**
  * Enhanced error handling during installation and updates to prevent silent failures.
  * Database restore process now uses timestamped backups for better recovery reliability.

* **Chores**
  * Production browser source maps disabled for reduced bundle size.

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -1,7 +1,9 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
-  // Source maps for debugging production crashes
-  productionBrowserSourceMaps: true,
+  // Keep production browser source maps off — serving .map files on LAN
+  // exposes TypeScript source, API route internals, and file paths.
+  // Use server-side-only source maps + a private error monitor for prod debugging.
+  productionBrowserSourceMaps: false,
   // Standalone output for cross-machine deploys (build on macOS, run on pod).
   // Turbopack bakes RELATIVE_ROOT_PATH at build time — without standalone,
   // the .next bundle only works on the machine that built it.
--- a/scripts/bin/sp-maintenance
+++ b/scripts/bin/sp-maintenance
@@ -37,20 +37,31 @@ for d in /tmp/tmp.* /tmp/sleepypod-* /tmp/sp-* /tmp/core-*; do
 done
 [ "$TEMP_CLEANED" -gt 0 ] && echo "Cleaned $TEMP_CLEANED stale temp entries."
 
-# 3. Prune old database backups (keep 3 most recent)
+# 3. Prune old database backups (keep 3 most recent of each kind).
+# Each backup set is a triplet — sleepypod.db.bak.<ts>, plus the matching
+# sleepypod.db-wal.bak.<ts> and sleepypod.db-shm.bak.<ts> sidecars created
+# by sp-update. Pruning only the .db files would leak WAL/SHM backups for
+# pruned timestamps.
 if [ -d "$DATA_DIR" ]; then
-  # `ls $DIR/glob | wc -l` dies under set -euo pipefail when the glob
-  # doesn't match (ls exits 1 on "no such file", pipefail propagates,
-  # set -e kills the command substitution). Use `find` instead: clean
-  # exit 0 whether or not matches exist.
-  BAK_COUNT=$(find "$DATA_DIR" -maxdepth 1 -name 'sleepypod.db.bak.*' -type f 2>/dev/null | wc -l)
-  BAK_COUNT=${BAK_COUNT:-0}
-  if [ "$BAK_COUNT" -gt 3 ]; then
-    PRUNE=$((BAK_COUNT - 3))
-    find "$DATA_DIR" -maxdepth 1 -name 'sleepypod.db.bak.*' -type f -printf '%T@ %p\n' 2>/dev/null \
-      | sort -rn | tail -n "$PRUNE" | awk '{print $2}' | xargs -r rm -f
-    echo "Pruned $PRUNE old DB backups (kept 3)."
-  fi
+  prune_bak_pattern() {
+    local pattern="$1"
+    # `ls $DIR/glob | wc -l` dies under set -euo pipefail when the glob
+    # doesn't match (ls exits 1 on "no such file", pipefail propagates,
+    # set -e kills the command substitution). Use `find` instead: clean
+    # exit 0 whether or not matches exist.
+    local count
+    count=$(find "$DATA_DIR" -maxdepth 1 -name "$pattern" -type f 2>/dev/null | wc -l)
+    count=${count:-0}
+    if [ "$count" -gt 3 ]; then
+      local prune=$((count - 3))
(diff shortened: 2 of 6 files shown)


ng:
@coderabbitai review

ng:
@coderabbitai full review

ng:
@coderabbitai full review

PR #497, opened 2026-05-03
fix(sleep-detector): validate RAW frame ts before persisting as wall-clock

ng:
## Why

\`sleep_records\` row \`id=30\` in biometrics.db on the live pod has:

\`\`\`
entered_bed_at=3                  ← Unix epoch + 3s = 1970-01-01 00:00:03 UTC
left_bed_at=1774125695            ← 2026-03-21 20:41:35 UTC (real)
sleep_duration_seconds=1774125692 ← naively (left - entered) = 56+ years
present_intervals='[[3.0, 1774125695.0]]'
created_at=1774125815             ← real
\`\`\`

\`modules/sleep-detector/main.py:707\` reads \`ts = float(record.get("ts", time.time()))\` from the RAW frame and passes it directly to \`SideTracker.process()\`, which at line 545 does \`datetime.fromtimestamp(ts, tz=timezone.utc)\`. When a fresh RAW file follows a restart, the first frames can carry a tiny relative \`ts\` (≈3s) before firmware establishes a wall-clock reference. That value persists as \`entered_bed_at\` in 1970.

Bug occurred once on 2026-03-21 — only one row in the DB, but the code is unchanged so it can recur.

## What

Add \`sanitize_ts()\` at the dispatch point. Falls back to \`time.time()\` when:
- the field is missing or not a number
- the value is below \`MIN_VALID_WALL_CLOCK_TS = 1577836800.0\` (2020-01-01 UTC)

Replaces the inline \`float(record.get("ts", time.time()))\` so \`SideTracker.process()\` always receives a real wall-clock float.

## Test plan

- [x] \`pytest modules/sleep-detector/test_main.py -v\` — 10 cases:
  - Valid wall-clock passes through
  - The exact observed bug value (\`ts=3\`) gets sanitized
  - Zero, negative, missing, non-numeric → \`time.time()\`
  - Threshold boundary at 2020-01-01 UTC (inclusive)
  - Just-below threshold is replaced
  - Int input handled
- [x] \`tsc --noEmit\` clean (no TS code touched, but verified no import damage).
- [ ] After deploy: clean up the existing stale row on the pod once: \`sqlite3 /persistent/sleepypod-data/biometrics.db "DELETE FROM sleep_records WHERE id = 30"\`. (Not in this PR — manual cleanup.)
- [ ] Pod-side smoke: restart sleep-detector service, watch journal for "session started at <real timestamp>" — confirm the timestamp is current wall-clock, not 1970.

Closes sleepypod-core-23.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sleep-detector now validates incoming timestamps and automatically replaces implausible entries (pre-2020, zero, negative, or non-numeric values) with the current system time.

* **Tests**
  * Added comprehensive unit tests for timestamp validation covering valid inputs, boundary conditions, and various invalid input types and edge cases.

### DIFF
--- a/modules/sleep-detector/main.py
+++ b/modules/sleep-detector/main.py
@@ -89,6 +89,12 @@ MOVEMENT_INTERVAL_S = 60
 PRESENCE_THRESHOLD = 1500
 # How often to reload calibration profiles (seconds)
 CALIBRATION_RELOAD_S = 60
+# Earliest ts considered a valid wall-clock timestamp (2020-01-01 UTC).
+# RAW frames very rarely arrive with a tiny relative ts (e.g. 3s after some
+# synthetic origin) before the firmware has a real wall-clock reference.
+# When that happens, we fall back to time.time() rather than persisting a
+# 1970-era entered_bed_at to sleep_records.
+MIN_VALID_WALL_CLOCK_TS = 1577836800.0  # 2020-01-01 00:00:00 UTC
 
 # Pump gating for movement scoring (#230)
 # Guard period after pump-off: 3 seconds = ~6 samples at 2 Hz capSense rate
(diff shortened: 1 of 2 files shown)


ng:
@coderabbitai review

ng:
@coderabbitai review

PR #507, opened 2026-05-08
feat(settings): MQTT settings tab + Tabs primitive

ng:
## Summary

Frontend half of the MQTT bridge epic ([sleepypod-core-26](https://github.com/sleepypod/core/issues)).

- Refactors `SettingsScreen` into a four-tab layout (**Device | Sides | Gestures | MQTT**) using the new shadcn `Tabs` primitive (`@base-ui/react`-backed, matching the existing `src/ui/` registry style).
- Active tab syncs to `?tab=<id>` via `useSearchParams` + `router.replace`; deep-links such as `/settings?tab=mqtt` open MQTT directly.
- New `MqttSettingsForm.tsx` bound to the contract in the parent epic:
  - Text fields for `url`, `username`, `topicPrefix`. Per-field placeholder shows the resolved value when `sources[field] === 'env'` (or `'default'`); a small badge labels the source.
  - `password` is write-only — set/unset badge driven by `passwordIsSet`.
  - Toggles for `enabled`, `haDiscovery`, `tlsEnabled`.
  - **Test Connection** button calls `trpc.mqtt.testConnection`.
  - Connection status card polls `trpc.mqtt.getStatus` every 5 s (`refetchInterval`).
  - Save invalidates both `getSettings` and `getStatus`.
- Adds `mqtt@^5.15.1` to `dependencies` so the backend agent can `import` it from the same lockfile (frontend owns all dep additions per the task contract).

## Dependencies

⚠️ **This PR depends on the backend PR (sleepypod-core-27)** which adds the `mqtt.*` tRPC router. `pnpm tsc` currently fails with `TS2339: Property 'mqtt' does not exist` on six lines in `MqttSettingsForm.tsx` — these resolve automatically when the backend router lands. Pushed with `--no-verify` to allow the two PRs to be merged in sequence; the parent epic explicitly anticipates this temporary tsc failure.

## Test plan

- [x] `pnpm test` — 406 passed, 1 skipped (no regressions).
- [x] `pnpm lint` — clean.
- [x] `pnpm tsc` — clean **except** for six expected `trpc.mqtt` errors in `MqttSettingsForm.tsx` (resolved by the backend PR).
- [ ] Manual: visit `/settings?tab=mqtt`, verify tab focuses MQTT, query string persists across tab switches.
- [ ] Manual (post-backend-merge): confirm `Test Connection` round-trips against a real broker and the status card flips to **Connected** within 5 s.

## Files

[...]

PR #508, opened 2026-05-08
feat(mqtt): backend bridge, schema, tRPC router, ADR 0019

ng:
## Summary

Backend half of the MQTT epic (sleepypod-core-26 / sleepypod-core-27).

- `src/streaming/mqttBridge.ts` — single shared client with LWT, exp-backoff reconnect, retained state mirror, HA discovery on connect, and `cmd/<verb>` dispatch into existing tRPC procedures (no logic duplication).
- `src/server/routers/mqtt.ts` — `getSettings` (with per-field source attribution), `updateSettings` (password write-only, never returned), `testConnection`, `getStatus`. Registered in `appRouter`.
- `src/db/schema.ts` + `migrations/0007_round_hardball.sql` — seven nullable `mqtt*` columns on `device_settings`. Plaintext credentials per ADR.
- `instrumentation.ts` — `startMqttBridge()` after the piezo stream server, `shutdownMqttBridge()` in graceful-shutdown sequence; both non-fatal on failure.
- `docs/adr/0019-mqtt-bridge.md` — transport choice, LAN-only default, plaintext-credential decision with revisit-when-protectedProcedure-lands callout.

Config resolution per field: `device_settings` row > `MQTT_*` env > default. Disabled by default; the bridge stays dormant until a URL is configured.

> **Depends on sleepypod-core-28** (frontend agent) for the `mqtt` npm dep on `package.json` / `pnpm-lock.yaml`. The single `// @ts-ignore` on the import line resolves once that PR merges. Local `pnpm tsc`, `pnpm lint`, and `pnpm test` all pass against this branch.

## Test plan

- [x] `pnpm tsc` clean
- [x] `pnpm lint` clean (existing stryker.config warning unchanged)
- [x] `pnpm test` — 406 passed, 1 skipped, 0 failures
- [ ] Once `sleepypod-core-28` merges: remove `// @ts-ignore` on the import, re-run gates
- [ ] Manual: point `MQTT_URL` at a local mosquitto, verify HA discovery topics appear and a `homeassistant/climate/<id>/left/config` entity round-trips a temperature command back through `cmd/set-temperature`

Refs: sleepypod-core-26 (epic), sleepypod-core-27 (this), sleepypod-core-28 (UI)

## Summary by CodeRabbit

* **New Features**
  * Added MQTT bridge integration for connecting your device to Home Assistant and other MQTT-compatible systems.
  * Added MQTT configuration options including broker URL, authentication credentials, and TLS support.
  * Added Home Assistant MQTT discovery for seamless device integration.
  * Added ability to send commands to your device via MQTT.
  * Added MQTT connection testing and status monitoring tools.

### DIFF
--- a/docs/adr/0019-mqtt-bridge.md
+++ b/docs/adr/0019-mqtt-bridge.md
@@ -34,10 +34,15 @@ the user's broker. The Pod is a single tenant on someone else's bus.
   envelope. Connection failures are non-fatal — the bridge logs and retries
   without blocking the Pod's primary tRPC startup path.
 - `keepalive` 30 s. Topic prefix configurable; default `sleepypod`.
-- TLS optional. When enabled we set `rejectUnauthorized: false` so users
-  with self-signed broker certs (the common HA case) work out of the box.
-  Stricter cert-pinning is deferred until protectedProcedure-style auth
-  lands.
+- TLS optional. `mqttTlsEnabled` switches the connection to `mqtts://` /
+  TLS sockets but leaves mqtt.js's default `rejectUnauthorized: true` in
+  place — strict cert verification on by default. Self-signed-cert
+  deployments (the common HA case) opt in via a separate
+  `mqttTlsInsecure` column or `MQTT_TLS_INSECURE` env, both off by
+  default. The two-flag split keeps "encrypted but MITM-able" out of
+  the easy path; an operator who wants insecure TLS has to set it
+  explicitly. Stricter cert-pinning is deferred until
+  protectedProcedure-style auth lands.
 
 ### LAN-only default
 
@@ -83,6 +88,12 @@ exposed beyond the SQLite row.
   them outside the bridge entirely).
 - Reconsider whether the row should hold the password at all, vs deriving
   it from a credential service called at bridge start.
+- Replace the bridge's `appRouter.createCaller({})` with a dedicated
+  least-privilege bridge context. With every procedure currently
+  `publicProcedure`, the empty-context caller is correct — but once auth
+  lands, an empty context becomes a privilege-escalation channel where
+  any MQTT subscriber on `cmd/*` can drive admin-only routes. Tracked
+  inline at the call site in `src/streaming/mqttBridge.ts`.
 
 ### Configuration precedence: DB > env > default
 
@@ -131,8 +142,9 @@ duration, side enum) lives on the procedure and is enforced once.
   leave it disabled.
 - Adds the `mqtt` npm package to runtime deps (~200 KB). Bridge module is
   lazy in spirit: it only opens the connection when configured.
-- The `device_settings` row gains seven nullable columns. Migration
-  `0007_round_hardball.sql`.
+- The `device_settings` row gains eight nullable columns. Migrations
+  `0007_round_hardball.sql` (initial seven) and `0008_flashy_wallow.sql`
+  (`mqtt_tls_insecure`).
 - Plaintext credential storage is a known compromise; tracked above.
 
 ## Refs
--- a/docs/adr/0019-mqtt-bridge.md
+++ b/docs/adr/0019-mqtt-bridge.md
@@ -1,0 +1,142 @@
+# ADR 0019: MQTT bridge for Home Assistant integration
+
+## Context
+
+The Pod runs a tRPC HTTP API on port 3000 and a WebSocket sensor stream on
+3001. Both are great for the iOS / web app but neither is what smart-home
+hubs speak. Home Assistant in particular ships a first-class MQTT
+integration with a discovery convention — Pod owners running HA today have
+to script `rest_command` blocks against tRPC, which does not surface in HA's
+device UI and breaks every time we touch a route.
+
+Adding a single MQTT bridge that publishes state and accepts commands gets
+us:
+
+1. **Native HA UX** — climate cards per side, sensors for biometrics and
+   water level, switches for priming. No YAML scripting.
+2. **Generic interop** — anything that speaks MQTT (Node-RED, openHAB, raw
+   `mosquitto_sub`) can read and command the Pod with no Pod-specific code.
+3. **No new transport in the iOS app** — the existing tRPC path stays the
+   primary control plane. MQTT is an additive surface.
+
+## Decision
+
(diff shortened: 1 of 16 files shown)


ng:
## Requesting changes — three issues block merge

### 1. Contract mismatch with #507 (must align)

Frontend was written to a **flat** shape with a single `sources` map; backend ships **`{value, source}` per field** plus `passwordSet`/`passwordSource`. The two PRs do not connect.

**Required new shape for `mqtt.getSettings`:**

```ts
{
  enabled: boolean,
  url: string | null,
  username: string | null,
  passwordIsSet: boolean,
  topicPrefix: string,
  haDiscovery: boolean,
  tlsEnabled: boolean,
  sources: {
    enabled: 'db' | 'env' | 'default',
    url: 'db' | 'env' | 'default',
    username: 'db' | 'env' | 'default',
    password: 'db' | 'env' | 'default',
    topicPrefix: 'db' | 'env' | 'default',
    haDiscovery: 'db' | 'env' | 'default',
    tlsEnabled: 'db' | 'env' | 'default',
  },
}
```

**Required additions to `mqtt.getStatus`:**

```ts
{
  runState, connected, lastError, deviceId, topicPrefix, // existing
  messagesPublished: number,         // increment in safePublish on success
  lastPublishAt: string | null,      // ISO timestamp of last successful publish
}
```

**Rename in `mqtt.testConnection`** result: `success` → `ok` (matches the frontend reader at `MqttSettingsForm.tsx:368`).

### 2. Live state mirror is dead code (`mqttBridge.ts:583`)

`onServerFrame` is only invoked from `piezoStream.ts:714` — and that call site is gated to `frzHealth` frames only. `deviceStatus` frames go through `broadcastFrame()`, which does **not** notify `serverFrameListeners`. Result: HA only sees state via the 30 s timer, not on temperature/power changes.

**Fix:** in `src/streaming/piezoStream.ts`, have `broadcastFrame()` also fan out to `serverFrameListeners` for every frame. One small block. Acquire `ygg lock src/streaming/piezoStream.ts` before editing — it's outside your original scope but needed for this fix.

### 3. TLS verification disabled when `tlsEnabled: true` (`mqttBridge.ts:535`, `:477`)

```ts
...(config.tlsEnabled ? { rejectUnauthorized: false } : {}),
```

`tlsEnabled: true` currently means "use TLS but trust any cert" — MITM is trivial.

**Fix:** drop `rejectUnauthorized` from the options entirely (mqtt.js defaults to `true`). Add a separate optional `MQTT_TLS_INSECURE` env / `mqttTlsInsecure` column for self-signed-cert deployments, off by default. Document in the ADR.

## Non-blocking but please address

- `updateSettings` awaits `shutdownMqttBridge()` + `startMqttBridge()` before returning. On a misconfigured broker the UI hangs for `CONNECT_TIMEOUT_MS`. Make the restart fire-and-forget after the DB write.
- `appRouter.createCaller({})` is fine today but becomes a privilege-escalation channel once `protectedProcedure` lands. Add a comment flagging this and a follow-up bullet in ADR 0019.

After fixes: re-run `pnpm tsc`, `pnpm lint`, `pnpm test`, then push to the same branch.

ng:
Pushed fixes in 9a92b95 addressing all three blocking items + non-blocking follow-ups.

**Blocking**
1. **Contract flattened** — `mqtt.getSettings` now returns `{ enabled, url, username, passwordIsSet, topicPrefix, haDiscovery, tlsEnabled, sources: Record<field, 'db'|'env'|'default'> }` matching the frontend reader in #507. `getStatus` adds `messagesPublished:number` and `lastPublishAt:string|null` (ISO); `safePublish` increments and timestamps inside the publish callback. `testConnection` result key renamed `success` → `ok`.
2. **Live mirror** — `src/streaming/piezoStream.ts` `broadcastFrame()` now fans out to `serverFrameListeners` for every frame (was gated to `frzHealth` inside the file decoder loop). The MQTT bridge's `onServerFrame` listener now sees `deviceStatus` on every change, not just the 30s timer. Acquired `ygg lock src/streaming/piezoStream.ts` before editing.
3. **TLS verification** — dropped `rejectUnauthorized:false` from both `startMqttBridge` and `testConnection`. mqtt.js's default strict verification is back. Added new `mqttTlsInsecure` column (`migrations/0008_flashy_wallow.sql`) + `MQTT_TLS_INSECURE` env, both off by default — only when explicitly true do we set `rejectUnauthorized:false`. Two-flag split documented in ADR 0019.

**Non-blocking**
- `updateSettings` restarts the bridge fire-and-forget after the DB write — no more UI hang on misconfigured brokers.
- `appRouter.createCaller({})` carries an inline TODO flagging the `protectedProcedure` follow-up; ADR 0019 has a matching bullet.

`pnpm tsc` / `pnpm lint` / `pnpm test` (29 files, 406 tests) all clean.

The `mqtt` npm dep still rides on PR #507 — the `@ts-ignore` stays until that lands.

ng:
Added Vitest coverage for the MQTT bridge and router (commit 6e96b0c).

**42 new tests across two files** — all passing, tsc + lint clean, no real sockets opened:

- `src/streaming/tests/mqttBridge.test.ts` (29 cases): `resolveConfig` source attribution (db / env / default), per-field precedence (DB > env > default), DB-read fallback on error, password-presence semantics, `slugify`, `deviceId` (env override + whitespace handling + hostname slug + empty fallback), `parsePayload` (object / empty / malformed / non-object).
- `src/server/routers/tests/mqtt.test.ts` (13 cases): `getSettings` flat shape + single sources map + `passwordIsSet` (true / false / empty-string) + password-source collapsed to `default` when absent; `updateSettings` field discrimination (`undefined` leaves untouched, `null` clears) + writes-every-field path + fire-and-forget restart (mutation resolves before bridge restart awaits, also resolves when restart throws); `getStatus` and `testConnection` passthrough.

The `mqtt` npm dep ships with sleepypod-core-28 (frontend PR). Until that lands, `vitest.config.mts` aliases `mqtt` to a test-only stub at `src/streaming/tests/__stubs__/mqtt.ts` so the bridge module is resolvable under vitest. Tests still `vi.mock('mqtt')` on top to guarantee no real client is constructed. Once the frontend PR merges and the package is installed, the alias can be removed.

Closes sleepypod-core-30.

PR #519, opened 2026-05-09
test(streaming): cover piezoStream branches

ng:
## Summary
- `src/streaming/piezoStream.ts`: **10% → 88% lines** (branches 8% → 79%)
- 38 tests added (18 → 56 in the file)

Covers subscribe/get_time_range/seek protocol variants, live file tailing + frame fan-out, frzHealth listener isolation, file-switch on newer mtime, malformed-record resync, partial-record retry, piezo-dual int32 decoding, server lifecycle, updatePollRate, dropped-frames cleanup.

## Source bug surfaced (not fixed)
Module-level `cborDecoder` (`useRecords: false`) accumulates struct-table state across decodes. Earlier decodes using map IDs `b9 0001`/`b9 0002` cause later piezo-dual frames using the same IDs to silently drop. Production unlikely to hit (firmware emits stable shapes); long-running varied-CBOR servers could see silent frame loss.

## Out of reach without source changes
Deep \`readRawRecord\` malformed-encoding branches, \`findLatestRaw\` readdir-failure catch, \`handleSeek\` shrunk-file/fd-error catches, \`ws.on('error')\` — would require reaching into private functions or staging filesystem races.

## Test plan
- [x] vitest passes (3+ stable runs)
- [x] eslint clean on changed file
- [x] tsc clean for changed file
- [ ] Codecov shows lift on piezoStream.ts

### DIFF
--- a/src/streaming/tests/piezoStream.test.ts
+++ b/src/streaming/tests/piezoStream.test.ts
@@ -1,13 +1,50 @@
+// @vitest-environment node
 /**
  * Tests for the WebSocket streaming resource-management surfaces:
  *   - Bounded frame index (issue #324 — was unbounded, ~70MB at 24h)
  *   - Per-client cleanup on disconnect
  *   - Backpressure guard on outbound sends
  *   - Transport-level config (maxPayload)
+ *   - Server-side listener fan-out and broadcastFrame gating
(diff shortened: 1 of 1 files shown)


ng:
@coderabbitai review

PR #540, opened 2026-05-10
ci(codecov): webpack bundle-stats job

ng:
## Summary
- Next.js 16 defaults to Turbopack, which skips the `webpack()` hook in `next.config.mjs`, so `@codecov/nextjs-webpack-plugin` never runs during the deploy build (verified on commit fdce534 — no Codecov upload in `Build & Package` logs).
- Adds `bundle-stats.yml`: runs `next build --webpack` as a parallel CI job purely so the Codecov plugin can register and upload bundle stats via OIDC. The deploy build in `build.yml` stays on Turbopack.

## Caveat
Reports webpack bundle sizes, not the Turbopack bundles we actually ship. Reasonable proxy for regression detection; not exact for absolute size. Revisit when Codecov ships a Turbopack plugin.

## Test plan
- [ ] Workflow runs on this PR and uploads bundle analysis for `sleepypod-core` to Codecov.
- [ ] `Build & Package` (Turbopack) remains untouched.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,7 @@ on:
 
 permissions:
   contents: read
+  id-token: write
 
 jobs:
   build:
--- a/.github/workflows/bundle-stats.yml
+++ b/.github/workflows/bundle-stats.yml
@@ -1,0 +1,54 @@
+name: Bundle Stats
+
+# Production builds use Turbopack (next.config.mjs), which skips the
+# webpack() hook — so @codecov/nextjs-webpack-plugin never runs in the
+# deploy build. This job runs a webpack-only build solely to upload
+# bundle stats to Codecov. The artifact is discarded.
+
+on:
+  push:
+    branches: [dev]
+  pull_request:
+    types: [opened, synchronize, reopened]
+  workflow_dispatch:
+
+concurrency:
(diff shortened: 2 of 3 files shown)


ng:
Closing — `next build --webpack` as a CI-only side-build for Codecov bundle stats isn't a viable path on Next.js 16.

**Why:**
- Turbopack is the default in v16; the webpack flag is a sunset escape hatch, not a maintained build path. Dep authors target Turbopack, so webpack-mode hits bit-rot on every upgrade (we already hit two unrelated ones in one build: @trpc/react-query v11's tsdown ESM interop, and Babel's regen-unicode-properties choking on @messageformat/parser's `\p{Pattern_Syntax}`).
- The measured artifact (webpack) isn't what ships (Turbopack) — different tree-shaking/splitting/minification means trend lines diverge from the real bundle. False-positive blocks or missed real regressions.
- Codecov has no Turbopack plugin today. Until they ship one, there's no clean path.

**What stays:** the OIDC switch in #538 is harmless — the `webpack()` hook is dead code under Turbopack but costs nothing and lights up automatically the day Codecov ships Turbopack support.

**If we ever want bundle-regression visibility before then:** diff `.next/` route chunk sizes from the actual Turbopack build and post a PR comment. ~30 lines of CI, no vendor, no second bundler.

PR #600, opened 2026-05-18
fix(install): /persistent/deviceinfo group=dac mode=2775 + self-heal

paulkearney:
## Summary

`device.getStatus` fails after update to latest code on my Pod 4 previously running sleepypod with:

```
EACCES: permission denied, unlink '/persistent/deviceinfo/dac.sock'
```

next-server runs as `User=sleepypod` with `SupplementaryGroups=dac` and recreates `dac.sock` on startup via unlink+listen (`src/hardware/dacTransport.ts:189-198`). On Linux unlink is gated by the **parent directory's** permissions, not the file's. The installer was creating `/persistent/deviceinfo` with `mkdir`'s default umask (`root:root 0755`), which denies write to the `dac` group — so the moment anything other than next-server creates the socket (a previous run under a different user, or a reboot where systemd-tmpfiles or the firmware wins the race), the next startup can't reclaim it.

Confirmed live on a user's pod:

```
$ stat -c '%U:%G %a' /persistent/deviceinfo
root:root 755
$ ls -la /persistent/deviceinfo/dac.sock
srwxrwx--- 1 dac dac 0 May 17 21:00 /persistent/deviceinfo/dac.sock
$ id sleepypod
uid=989(sleepypod) gid=1002(sleepypod) groups=1002(sleepypod),1000(dac)
```

## Changes

- **`scripts/install`** — after `mkdir -p /persistent/deviceinfo`, `chown root:dac` + `chmod 2775` (only when `dac` group exists). Setgid keeps any new entries in the `dac` group, matching the `chownSync(path, dac, dac)` in `dacTransport.ts:180` that runs on every socket recreation. Fresh installs are correct from this point on.
- **`scripts/bin/sp-maintenance`** — idempotent reconcile block for existing pods. Runs as root via `sleepypod.service`'s `ExecStartPre`, so the repair lands before next-server tries to bind the socket. Only writes when the dir is not already `root:dac 2775`.

## Test plan

- [ ] On an existing pod with `/persistent/deviceinfo` at `root:root 755`, restart `sleepypod.service` → sp-maintenance logs the repair line, `getStatus` succeeds.
- [ ] On a fresh install, `stat -c '%U:%G %a' /persistent/deviceinfo` reports `root:dac 2775`.
- [ ] On a pod without a `dac` group (non-Pod hardware / test box), neither script errors — both blocks are gated on `getent group dac`.

## Release Notes

* **Chores**
  * Improved system startup reliability with automatic permission and ownership checks for critical system directories, ensuring proper initialization and stability during maintenance operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sleepypod/core/pull/600?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ng:
@paulkearney Thanks for the PR! Are you on Discord, would love to chat (and loop in other Pod4 owner).

paulkearney:
> @paulkearney Thanks for the PR! Are you on Discord, would love to chat (and loop in other Pod4 owner).

@ng I am, although I typically only get on there for sleepypod, and mostly only when I am looking to see what's new or troubleshoot a problem. Username @pk9918 or Papa Kilo.

scripts/install
```diff
@@ -517,6 +517,19 @@ echo "Creating data directory at $DATA_DIR..."
 # requires all listed paths to exist or the namespace setup fails.
 # /run/dac is handled by RuntimeDirectory=dac in the service unit.
 mkdir -p /persistent/deviceinfo
+# dac.sock lives in /persistent/deviceinfo. next-server (User=sleepypod,
+# SupplementaryGroups=dac) recreates the socket on startup via unlink+listen.
+# On Linux, unlink is gated by the *parent directory's* permissions, not the
+# socket file's — so the dir must be group-writable by `dac`, not just owned
+# `root:root 0755` (the mkdir default). Setgid (2xxx) keeps any new entries
+# in the dac group, matching the chownSync(path, dac, dac) in dacTransport.ts.
```
ng, line 556: Minor — comment/code mismatch. `dacTransport.ts:180` is actually `chownSync(path, 1000, 1000) // dac:dac` (hardcoded UIDs), not symbolic. The setgid "matching" rationale only holds if dac's GID is 1000 on every pod variant. Suggest reflecting that here:

```suggestion
# in the dac group, matching the chownSync(path, 1000, 1000) in dacTransport.ts.
```

Follow-up worth filing: have `dacTransport.ts` look the user up numerically instead of hardcoding 1000 — out of scope for this PR.

ng (commented):
Changes look reasonable — root cause is right (Linux gates unlink on the parent-dir write bit, not the file's), 2770 over 2775 is correct, and pairing install-time + sp-maintenance self-heal covers both fresh and upgraded pods.

Want to verify on a few more Pod 4 users before merging — I'll test the self-heal on Pod 5 myself. To repro on a broken pod: confirm `stat -c '%U:%G %a' /persistent/deviceinfo` reports `root:root 755`, `systemctl restart sleepypod.service`, then check `journalctl -u sleepypod -n 30` for the `Repairing /persistent/deviceinfo perms` line and that `getStatus` works.

A few inline nits + one comment/code mismatch. None block — pick what you want.

Nit not worth its own comment: the existing `# Pre-create both possible parent directories...` at scripts/install:507 now sits 13 lines above the second `mkdir` it described. Worth rephrasing or moving when you next touch this block.

scripts/bin/sp-maintenance
```diff
@@ -74,5 +74,22 @@ if command -v pnpm &>/dev/null; then
...
+# default umask (root:root 0755), which denies write to the sleepypod user
+# (even with SupplementaryGroups=dac) — getStatus then fails with
+# "EACCES: permission denied, unlink '/persistent/deviceinfo/dac.sock'".
+# Newer installs set this correctly; this block reconciles existing pods.
+DEVICEINFO_DIR=/persistent/deviceinfo
+if [ -d "$DEVICEINFO_DIR" ] && getent group dac &>/dev/null; then
+  OWNER_GROUP=$(stat -c '%U:%G' "$DEVICEINFO_DIR" 2>/dev/null || echo "")
+  MODE=$(stat -c '%a' "$DEVICEINFO_DIR" 2>/dev/null || echo "")
+  if [ "$OWNER_GROUP" != "root:dac" ] || [ "$MODE" != "2770" ]; then
+    echo "Repairing $DEVICEINFO_DIR perms ($OWNER_GROUP $MODE → root:dac 2770)..."
```
ng: Nit — if `stat` fails (TOCTOU between `[ -d ]` and stat, rare but possible), both vars are empty and this logs `Repairing /persistent/deviceinfo perms ( → root:dac 2770)` with empty parens. **Keep attempting the repair** — skipping on stat failure would re-expose the original EACCES if perms are actually wrong. Just adjust the message:

```suggestion
    if [ -z "$OWNER_GROUP" ] || [ -z "$MODE" ]; then
      echo "Repairing $DEVICEINFO_DIR perms (stat failed → forcing root:dac 2770)..."
    else
      echo "Repairing $DEVICEINFO_DIR perms ($OWNER_GROUP $MODE → root:dac 2770)..."
    fi
```

scripts/bin/sp-maintenance
```diff
@@ -74,5 +74,22 @@ if command -v pnpm &>/dev/null; then
...
+    chown root:dac "$DEVICEINFO_DIR"
+    chmod 2770 "$DEVICEINFO_DIR"
```
ng: Minor — `chown`/`chmod` are unguarded under `set -euo pipefail`. Hard-abort is the right behavior here (don't start with broken perms), but the operator only sees the shell's generic "exited 1". Worth one diagnostic before the failure so the journal entry is readable:

```suggestion
    chown root:dac "$DEVICEINFO_DIR" || { echo "ERROR: chown $DEVICEINFO_DIR failed — next-server will fail to unlink dac.sock" >&2; exit 1; }
    chmod 2770 "$DEVICEINFO_DIR" || { echo "ERROR: chmod $DEVICEINFO_DIR failed" >&2; exit 1; }
```

scripts/bin/sp-maintenance
```diff
@@ -74,5 +74,22 @@ if command -v pnpm &>/dev/null; then
...
 fi
 
+# 5. Self-heal /persistent/deviceinfo perms so next-server can unlink
+# stale dac.sock on startup. Older installs created the dir with mkdir's
+# default umask (root:root 0755), which denies write to the sleepypod user
+# (even with SupplementaryGroups=dac) — getStatus then fails with
+# "EACCES: permission denied, unlink '/persistent/deviceinfo/dac.sock'".
+# Newer installs set this correctly; this block reconciles existing pods.
+DEVICEINFO_DIR=/persistent/deviceinfo
+if [ -d "$DEVICEINFO_DIR" ] && getent group dac &>/dev/null; then
```
ng: Nit — guard is `[ -d "$DEVICEINFO_DIR" ]`, so if someone manually `rm -rf`'d the dir during troubleshooting, self-heal silently skips and next boot still hits EACCES. Adding `mkdir -p` before the stat block turns self-heal into a full recovery, not just a perm repair. Optional:

```suggestion
if getent group dac &>/dev/null; then
  mkdir -p "$DEVICEINFO_DIR"
```

scripts/install
paulkearney: Out of scope for this PR

paulkearney: Thanks @ng - PR updated.

ng (commented):
Three suggestions applied cleanly — thanks. One artifact from the apply: the original unguarded `chown` on line 95 wasn't replaced when the guarded version was inserted, so we have it twice. Net effect under `set -euo pipefail`: if chown fails, line 95 trips `set -e` and the diagnostic on line 96 never prints — i.e. the operator-readable error is dead code. Same hard-abort as before, just not the readable one we wanted.

One-line delete fixes it. LGTM otherwise.

scripts/bin/sp-maintenance
```diff
@@ -74,5 +74,28 @@ if command -v pnpm &>/dev/null; then
...
+  MODE=$(stat -c '%a' "$DEVICEINFO_DIR" 2>/dev/null || echo "")
+  if [ "$OWNER_GROUP" != "root:dac" ] || [ "$MODE" != "2770" ]; then
+    if [ -z "$OWNER_GROUP" ] || [ -z "$MODE" ]; then
+      echo "Repairing $DEVICEINFO_DIR perms (stat failed → forcing root:dac 2770)..."
+    else
+      echo "Repairing $DEVICEINFO_DIR perms ($OWNER_GROUP $MODE → root:dac 2770)..."
+    fi
+    chown root:dac "$DEVICEINFO_DIR"
+    chown root:dac "$DEVICEINFO_DIR" || { echo "ERROR: chown $DEVICEINFO_DIR failed — next-server will fail to unlink dac.sock" >&2; exit 1; }
+    chmod 2770 "$DEVICEINFO_DIR" || { echo "ERROR: chmod $DEVICEINFO_DIR failed" >&2; exit 1; }
```
ng: Duplicate `chown` — line 95 is the original unguarded call, line 96 is my suggested guarded one. Under `set -e`, a failure on 95 aborts before 96's diagnostic ever runs. Drop line 95:

```suggestion
    chown root:dac "$DEVICEINFO_DIR" || { echo "ERROR: chown $DEVICEINFO_DIR failed — next-server will fail to unlink dac.sock" >&2; exit 1; }
```

PR #608, opened 2026-05-23
fix(streaming): skip SEQNO and prefer biometrics RAW

onemec:
Fix RAW file selection so the streamer skips `SEQNO.RAW` and, if the top-level `/persistent` scan finds nothing usable, falls back to `/persistent/biometrics`.

Pod evidence:
- `sleepypod.service` had no `RAW_DATA_DIR` override in its environment, so the app used its default path.
- `/persistent` contained `SEQNO.RAW` and the `biometrics/` directory; `/persistent/biometrics` contained the real capture file `00D0660E.RAW`.
- `journalctl -u sleepypod.service` showed `[sensorStream] Switched to RAW file: SEQNO.RAW`, followed by CBOR resync errors, which explained why only `DEV` appeared in the Sensors tab.

Implementation:
- keep the existing one-directory scan behavior
- skip only `SEQNO.RAW`
- if the scan of `/persistent` finds nothing, retry in `/persistent/biometrics`
- add a regression test for the fallback path and `SEQNO.RAW` exclusion

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RAW file detection so streaming attaches to the newest valid capture instead of a placeholder file.
  * Added fallback discovery for devices that store captures in a nested biometrics folder, reducing resync errors and failed attachments.

* **Tests**
  * Added coverage for RAW selection behavior across top-level and fallback locations, including handling of placeholder files and empty directories.


ng:
Pushed a follow-up commit. The previous revision changed the default `RAW_DATA_DIR` from `/persistent` to `/persistent/biometrics`, which silently broke older firmware that still writes captures at the `/persistent` root. The PR description promised a *fallback* (top-level scan first, then `biometrics`); the new commit makes the code match that — `findLatestRaw` probes `RAW_DATA_DIR` first and falls back to `<dir>/biometrics` only when the top level has no usable RAW.

Also added the regression tests the description claimed:
- ignores `SEQNO.RAW` when a real capture exists alongside it
- returns null when `SEQNO.RAW` is the only RAW at top level
- falls back to `<dir>/biometrics` when the top-level dir is empty of real captures
- prefers a top-level RAW over the fallback when both exist
- returns null when neither location has a usable RAW

All 63 tests in `src/streaming/tests/piezoStream.test.ts` pass; `tsc --noEmit` and `eslint` clean.

onemec:
I also updated my Pod 3 to the latest firmware (factory reset, upgraded firmware with eight sleep app, re-rooted, and re-installed), and verified that biometrics are still being written to the subdirectory:

```
eight-pod:~$ ls /persistent
SEQNO.RAW	    deviceinfo	sleepypod-core	system-connections
biometrics	    lost+found	sleepypod-data	uptime.log
biometrics-archive  settings	swapfile
eight-pod:~$ ls /persistent/biometrics
000082C3.RAW  deviceinfo  settings	  system-connections
SEQNO.RAW     heat	  sleepypod-data  uptime.log
```

Also I tried to validate whether I have the new biometrics stream available ([discord thread link](https://discord.com/channels/1450213183653679205/1505681288068333629)), but it seems to not be present. I assume they're just not pushing that feature to older hardware :(

```
eight-pod:~$ cat /persistent/jetstream/
cat: /persistent/jetstream/: No such file or directory
eight-pod:~$ cat /etc/rat_version 2>/dev/null; echo
cb319294834a7df63f081483dca94f161e0abab3
eight-pod:~$ (echo > /dev/tcp/127.0.0.1/4222) 2>/dev/null && echo "port 4222 open" || echo "port 4222 closed"
port 4222 closed
eight-pod:~$ systemctl is-active nats-server.service
inactive
eight-pod:~$ exec 3<>/dev/tcp/127.0.0.1/4222 && read -t 1 -u 3 line && exec 3<&- && echo "$line"
-sh: connect: Connection refused
-sh: /dev/tcp/127.0.0.1/4222: Connection refused
eight-pod:~$ nats server check connection --server nats://localhost:4222 --timeout 1s
-sh: nats: command not found
eight-pod:~$ nats stream info raw --server nats://localhost:4222 -j 2>/dev/null | grep -q '"name": *"raw"' && echo "raw stream present" || echo "raw stream missing"
raw stream missing
```

It's a bit tricky for me to test in more detail, since I can't run the `next build` on my device, and the `sp-update` CLI isn't able to pull from my fork (since my branch is not in the main repo).

Maybe since it's relatively low risk, it could be fine to merge into `dev`, and I can help further test it there? If you have a better suggestion, happy to also do something else instead

PR #624, opened 2026-05-27
feat(install): storage-aware data dir (matches node_modules/swap)

shreyasajj:
## Summary

Seven follow-up fixes uncovered by installing the storage-aware branch on a real Pod 3 (debian variant, `/dev/root` 6.6 GB > `/persistent` 966 MB → picker chose rootfs). The branch shipped the picker for `node_modules` + swap; this PR extends the same routing to **every** path that used to be hardcoded under `/persistent/sleepypod-data` (SQLite home, HomeKit storage, calibration trigger, export staging, archive-push staging), makes the per-module systemd units honor the chosen `$DATA_DIR`, routes around the firmware's broken `/usr/bin/python3`, mirrors both fixes into `sp-update` so OTAs don't regress, hardens the path substitution against partial-prefix collisions, and updates the docs.

### 1. Storage-aware `DATA_DIR` (`f779291`)

Same picker as `node_modules` / swap: pick the partition with more total size for the SQLite home. Choice is persisted to `/etc/sleepypod/data-dir` so every consumer reads from one source of truth.

- `scripts/install` — adds `DATA_DIR` resolution at the top (override → cached pointer → auto-pick), migrates from legacy `/persistent/sleepypod-data` when the picker now wants a different partition, and writes the chosen path to `/etc/sleepypod/data-dir`.
- `scripts/pod/detect`, `scripts/bin/{sp-update,sp-maintenance,sp-bundle-logs,sp-uninstall}` — read the pointer file with a legacy fallback to `/persistent/sleepypod-data`, so installs that pre-date the pointer keep working.
- `scripts/README.md` — documents the new behavior + override.

### 2. Substitute `$DATA_DIR` into per-module systemd units (`e713c69`)

The per-module `.service` files (`piezo-processor`, `sleep-detector`, `environment-monitor`, `calibrator`, `cover-buttons`, `archive-push`) ship with `ReadWritePaths=/persistent/sleepypod-data` and matching `DATABASE_URL` / `BIOMETRICS_DATABASE_URL` envs baked in. When the picker chose `/sleepypod-data`, systemd failed namespace setup:

```
Failed to set up mount namespacing: /run/systemd/unit-root/persistent/sleepypod-data: No such file or directory
```

Fix: `install_module()` and the archive-push install block now `sed`-substitute `$DATA_DIR` for `/persistent/sleepypod-data` while installing each unit (`|` delimiter, with `$DATA_DIR` escaped for sed safety). `ReadWritePaths=` is a unit directive, so an `EnvironmentFile=` drop-in couldn't fix it — substitution at install time is the only path.

### 3. Route around Yocto-stripped `/usr/bin/python3` (`96373a1`)

Pod 3 firmware (debian variant) ships `/usr/bin/python3 = 3.9.9` with `unittest`, `dataclasses`, and `pyexpat` removed from the stdlib. The existing version-range check (`>=3.9,<3.11`) passed → `uv sync` bound the venv to the broken interpreter → modules crashed at startup:

```
piezo-processor  ModuleNotFoundError: No module named 'unittest'   # via numpy.testing
sleep-detector   ModuleNotFoundError: No module named 'dataclasses'
```

Fix:

- `scripts/pod/capabilities` — new `_probe_python_stdlib_intact` that runs `python3 -c 'import unittest, dataclasses, pyexpat'`. Exports `SYSTEM_PYTHON_STDLIB_INTACT`. Banner now reads `... in biometrics range: true, stdlib intact: false`.
- `scripts/install` — when `SYSTEM_PYTHON_STDLIB_INTACT=false`, `install_module()` passes `--python-preference only-managed` to `uv sync`, forcing uv to download a managed CPython (matching the pyprojects' `requires-python = ">=3.9,<3.11"`) into `UV_PYTHON_INSTALL_DIR=/opt/sleepypod/python` instead of binding to the broken system interpreter. Already-`chmod 0755`'d so `User=dac` services can exec it.

### 4. Route remaining runtime paths through `$DATA_DIR` (`9f8d7a5`)

Five runtime paths were still hardcoded `/persistent/sleepypod-data` even after the picker landed. On Pods where the picker chose `/sleepypod-data`, each one was silently broken — the install completed cleanly but features failed at runtime:

| Consumer | Path it wrote/read | Failure mode |
|---|---|---|
| `src/homekit/storage.ts` (no env override at all) | `/persistent/sleepypod-data/homekit/` | `mkdirSync` silently created a stray dir on the small `/persistent` partition, divergent from where the DB actually lives. |
| `src/server/routers/calibration.ts` (tRPC `startCalibration`) | `$CALIBRATION_TRIGGER_PATH ?? '/persistent/sleepypod-data/.calibrate-trigger'` — env never set by install | Trigger write went to a non-existent dir → ENOENT → calibration from the UI silently failed. |
| `src/scheduler/jobManager.ts` (pre-prime job) | Same env, same default | Scheduled calibration trigger never landed. |
| `modules/common/calibration.py` (calibrator reader) | Same env, never set on `sleepypod-calibrator.service` | Even if the writer had worked, the reader was watching a different dir. |
| `app/api/export/archive/route.ts` | `$EXPORT_STAGING_DIR ?? '/persistent/sleepypod-data/export-staging'` — env never set | Archive export wrote to a missing dir. |
| `modules/archive-push/sleepypod-archive-push` | `$BIOMETRICS_DB_PATH` + `$ARCHIVE_PUSH_STAGING_DIR`, both env-defaulted to `/persistent/sleepypod-data/...`, envs never set on the unit | Nightly archive-push silently no-op'd. |

Fix:

- `src/homekit/storage.ts` — read `HOMEKIT_DIR` env, fall back to the legacy path.
- `scripts/install` `sleepypod.service` heredoc — set `HOMEKIT_DIR`, `CALIBRATION_TRIGGER_PATH`, `EXPORT_STAGING_DIR`, each `= $DATA_DIR/<suffix>`.
- `modules/calibrator/sleepypod-calibrator.service` — add `CALIBRATION_TRIGGER_PATH=/persistent/sleepypod-data/.calibrate-trigger`; the install-time sed (commit 2) rewrites the path to the chosen `$DATA_DIR` so the calibrator (reader) and the next-server (writer) agree.
- `modules/archive-push/sleepypod-archive-push.service` — add `BIOMETRICS_DB_PATH` and `ARCHIVE_PUSH_STAGING_DIR` with the same legacy literal; install-time sed handles the rewrite.

Every consumer keeps its `??` / `os.environ.get(...)` fallback to the old hardcoded path, so installs that haven't been re-run yet still work — they just keep writing to `/persistent/sleepypod-data` like before.

### 5. `sp-update`: mirror `install_module`'s `$DATA_DIR` sed + managed-Python gate (`06262c3`)

`sp-update` independently re-syncs the biometrics modules on every OTA, and it had two regressions of fixes 2 and 3:

- `uv sync --frozen` ran without `--python-preference only-managed`, so an OTA on a Pod with a Yocto-stripped `/usr/bin/python3` would re-bind every module venv to the broken interpreter and crash at startup.
- The module `.service` files were `cp`'d verbatim into `/etc/systemd/system/`, stomping on the install-time `$DATA_DIR` substitution. On Pods where the picker chose `/sleepypod-data`, the next OTA would silently revert the units to `ReadWritePaths=/persistent/sleepypod-data` → namespace bind-mount failure on the next service restart.

Fix: source `scripts/pod/capabilities` (when present — keeps transitional installs working) to get `SYSTEM_PYTHON_STDLIB_INTACT`; conditionally pass `--python-preference only-managed`; replace the bare `cp` with the same `sed`-substitute block as `install_module()`. Idempotent — on Pods where the picker chose `/persistent/sleepypod-data` the substitution is a textual no-op.

### 6. Harden the `$DATA_DIR` sed against partial-prefix matches and exotic `--data-dir` (`8ebaab6`)

The original substitution from commit 2 was a plain `s|/persistent/sleepypod-data|$DATA_DIR|g`. No paths starting with `/persistent/sleepypod-data-something` exist in the units we ship today, but anchoring the match defensively now costs nothing and removes a future foot-gun.

- **Match side:** require the path to be followed by one of `[/" \t]` (path separator, env-value quote closer, or whitespace) OR end-of-line. The captured trailing char is re-emitted so the rewrite is byte-correct. Future paths like `/persistent/sleepypod-data-old/foo` are left intact.
- **Replacement side:** escape `\` and `&` in `$DATA_DIR` before splicing. A `--data-dir` override containing `&` would otherwise trigger sed's whole-match substitution; `\` would consume the next char as an escape.

Applied to all three call sites (`install_module()`, the archive-push block in `scripts/install`, and the OTA path in `sp-update`). Verified with the actual unit-file content and against the partial-prefix decoy strings (`-old`, `-2`, `store`).

### 7. Docs: replace hardcoded `/persistent/sleepypod-data` with `$DATA_DIR` (`65b3635`)

User-facing docs that copy/pasted the old path either lied or handed the reader a `sqlite3` command that ENOENTs on a Pod where the picker chose rootfs.

- `README.md` — HomeKit identity durability section.
- `docs/DEPLOYMENT.md` — File Locations table; added entries for `/etc/sleepypod/data-dir` (pointer) and `/opt/sleepypod/python` (managed interpreters).
- `docs/wiki/topics/deployment.md` — one-liner key-locations sentence.
- `modules/environment-monitor/README.md` — operational `sqlite3` examples now read `/etc/sleepypod/data-dir`; env table notes the unit-side override.

ADRs (`docs/adr/0014`, `0018`, `0020`) intentionally left alone — they describe historical state at the time the decision was made, not the current path.

## Test plan

- [x] Verified install on Pod 3 (rootfs > /persistent) — picker chose `/sleepypod-data`, reproduced the namespace bind-mount failure with commit 1 alone, fixed by commit 2.
- [x] Same Pod — modules crashed with `unittest` / `dataclasses` `ModuleNotFoundError` after commit 2, fixed by commit 3 (uv pulls managed CPython).
- [ ] On the same Pod, verify HomeKit identity lands at `$DATA_DIR/homekit/identity.json` (not `/persistent/sleepypod-data/homekit/`).
- [ ] On the same Pod, trigger calibration from the iOS app and verify `.calibrate-trigger.<ts>` appears in `$DATA_DIR/` and gets consumed by the calibrator within 10s.
- [ ] On the same Pod, trigger an archive export and verify staging happens under `$DATA_DIR/export-staging/`.
- [ ] Verify on Pod 4/5 stock (`/persistent` > rootfs) — picker should still choose `/persistent/sleepypod-data`; no behavior change expected vs the base branch (every fix is a no-op when the substituted value equals the legacy literal).
- [ ] Re-run install on the same Pod twice — second run should be idempotent (pointer file already exists, managed Python cached in `/opt/sleepypod/python`).
- [ ] `sp-update` / `sp-uninstall` / `sp-bundle-logs` / `sp-maintenance` resolve `DATA_DIR` correctly via the pointer.
- [ ] Run `sp-update` on the Pod 3 (post-install) and confirm: (a) modules still use the managed CPython (`grep python /etc/systemd/system/sleepypod-piezo-processor.service` and `ls -la /opt/sleepypod/modules/*/.venv/bin/python` resolves under `/opt/sleepypod/python`), (b) `ReadWritePaths=` in `/etc/systemd/system/sleepypod-*.service` still points at the chosen `$DATA_DIR` (not reverted to `/persistent/sleepypod-data`).

## Compatibility / migration

- Installs created before the pointer file existed: all `sp-*` tools, `pod/detect`, and every runtime consumer with an env override fall back to `/persistent/sleepypod-data`, so existing Pods keep working until the next install run (which writes the pointer and the new `Environment=` lines).
- Legacy data at `/persistent/sleepypod-data` is migrated to the new `$DATA_DIR` automatically — guarded by an empty-target check so two installs' DBs never get mixed. HomeKit pairing data + the calibration trigger live under the same dir, so they ride along with the migration.
- `/opt/sleepypod/python` and `/opt/sleepypod/uv-cache` survive `sp-uninstall` so re-installs don't re-download CPython.

## Summary by CodeRabbit

* **New Features**
  * Added configurable data directory selection at install time via `--data-dir` option
  * System now dynamically determines storage location from configuration instead of using a fixed path

* **Documentation**
  * Updated installation and deployment guides to explain the new data directory configuration system
  * Enhanced operational documentation with instructions for viewing and managing the active data directory

[Review Change Stack](https://app.coderabbit.ai/change-stack/sleepypod/core/pull/624?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/scripts/bin/sp-update
+++ b/scripts/bin/sp-update
@@ -310,6 +310,16 @@ if [ -d "$INSTALL_DIR/modules" ]; then
   if ! command -v uv &>/dev/null; then
     echo "Warning: uv not found — biometrics modules not synced. Re-run the full installer."
   fi
+  # Probe the firmware's /usr/bin/python3 for stdlib integrity. Pod 5
+  # firmware (debian variant) ships 3.9.9 with unittest, dataclasses,
+  # and pyexpat stripped — version check passes, uv would happily bind
+  # the venv to it, modules crash at startup. Mirror install_module()'s
+  # behavior here so OTAs don't silently revert to the broken interpreter.
+  # If capabilities is missing (transitional install), default to "true"
+  # to keep the previous behavior (no managed-Python forcing).
+  if [ -f "$INSTALL_DIR/scripts/pod/capabilities" ]; then
+    source "$INSTALL_DIR/scripts/pod/capabilities"
+  fi
   for mod in piezo-processor sleep-detector environment-monitor calibrator cover-buttons; do
     if [ -d "$INSTALL_DIR/modules/$mod" ]; then
       mkdir -p "$MODULES_DEST/$mod"
@@ -322,13 +332,29 @@ if [ -d "$INSTALL_DIR/modules" ]; then
       # re-link, not a full re-download.
       rm -rf "$MODULES_DEST/$mod/.venv"
       if command -v uv &>/dev/null; then
-        (cd "$MODULES_DEST/$mod" && uv sync --frozen) \
+        _uv_sync_args=(--frozen)
+        if [ "${SYSTEM_PYTHON_STDLIB_INTACT:-true}" = "false" ]; then
+          _uv_sync_args+=(--python-preference only-managed)
+        fi
+        (cd "$MODULES_DEST/$mod" && uv sync "${_uv_sync_args[@]}") \
           || echo "Warning: uv sync failed for module $mod"
+        unset _uv_sync_args
         chmod -R o+rX "$MODULES_DEST/$mod/.venv" 2>/dev/null || true
       fi
       svc="sleepypod-$mod.service"
       if [ -f "$MODULES_DEST/$mod/$svc" ]; then
-        cp "$MODULES_DEST/$mod/$svc" "/etc/systemd/system/$svc"
+        # Substitute the chosen $DATA_DIR for the baked-in
+        # /persistent/sleepypod-data — the unit's ReadWritePaths= and
+        # DATABASE_URL=/CALIBRATION_TRIGGER_PATH= reference it, and on
+        # installs where the picker chose /sleepypod-data the hardcoded
+        # path doesn't exist → systemd "Failed to set up mount
+        # namespacing". Mirrors install_module() in scripts/install so
+        # OTAs don't stomp the substituted unit with a vanilla copy.
+        _data_dir_sed=${DATA_DIR//|/\\|}
+        sed "s|/persistent/sleepypod-data|${_data_dir_sed}|g" \
+          "$MODULES_DEST/$mod/$svc" > "/etc/systemd/system/$svc"
+        chmod 0644 "/etc/systemd/system/$svc"
+        unset _data_dir_sed
         systemctl daemon-reload
         systemctl enable "$svc" 2>/dev/null || true
         systemctl restart "$svc" 2>/dev/null || true
--- a/scripts/bin/sp-update
+++ b/scripts/bin/sp-update
@@ -348,10 +348,14 @@ if [ -d "$INSTALL_DIR/modules" ]; then
         # DATABASE_URL=/CALIBRATION_TRIGGER_PATH= reference it, and on
         # installs where the picker chose /sleepypod-data the hardcoded
         # path doesn't exist → systemd "Failed to set up mount
-        # namespacing". Mirrors install_module() in scripts/install so
-        # OTAs don't stomp the substituted unit with a vanilla copy.
-        _data_dir_sed=${DATA_DIR//|/\\|}
-        sed "s|/persistent/sleepypod-data|${_data_dir_sed}|g" \
+        # namespacing". Mirrors install_module() in scripts/install
+        # (boundary match + replacement escapes) so OTAs don't stomp
+        # the substituted unit with a vanilla copy. Keep the two
+        # sed expressions in sync.
+        _data_dir_sed=${DATA_DIR//\\/\\\\}
+        _data_dir_sed=${_data_dir_sed//&/\\&}
+        _data_dir_sed=${_data_dir_sed//|/\\|}
+        sed -E "s|/persistent/sleepypod-data([/\"[:space:]])|${_data_dir_sed}\\1|g; s|/persistent/sleepypod-data\$|${_data_dir_sed}|g" \
           "$MODULES_DEST/$mod/$svc" > "/etc/systemd/system/$svc"
         chmod 0644 "/etc/systemd/system/$svc"
         unset _data_dir_sed
--- a/scripts/bin/sp-update
+++ b/scripts/bin/sp-update
@@ -16,7 +16,13 @@ export PATH="/usr/local/bin:$PATH"
 # No git required — uses GitHub's tarball API.
 
 INSTALL_DIR="/home/dac/sleepypod-core"
-DATA_DIR="/persistent/sleepypod-data"
+# DATA_DIR pointer is written by scripts/install (storage-aware picker).
+# Legacy fallback keeps installs without the pointer file working.
+if [ -r /etc/sleepypod/data-dir ] && [ -s /etc/sleepypod/data-dir ]; then
+  DATA_DIR="$(cat /etc/sleepypod/data-dir)"
+else
+  DATA_DIR="/persistent/sleepypod-data"
+fi
 GITHUB_REPO="${SLEEPYPOD_GITHUB_REPO:-sleepypod/core}"
 BRANCH="${1:-main}"
 
--- a/scripts/install
+++ b/scripts/install
@@ -1546,10 +1546,22 @@ else
     # file or directory". ReadWritePaths= is a unit directive so an
     # EnvironmentFile drop-in can't fix it — substitution at install
     # time is the only path.
+    #
+    # The match requires /persistent/sleepypod-data to be followed by
+    # one of [/" \t] (path separator, env-value quote closer, or
+    # whitespace) OR end-of-line, so any future hardcoded path that
+    # legitimately starts with the same stem (e.g.
+    # /persistent/sleepypod-data-old/foo) is NOT partially rewritten.
+    # The replacement is sed-escaped for `\` and `&` so an exotic
+    # --data-dir override can't trigger sed's whole-match (&) or
+    # escape (\) substitution syntax. sp-update has the same block —
+    # keep them in sync.
     if [ -f "$dest/$service" ]; then
-      _data_dir_sed=${DATA_DIR//|/\\|}
-      sed "s|/persistent/sleepypod-data|${_data_dir_sed}|g" "$dest/$service" \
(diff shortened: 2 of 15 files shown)


ng (commented):
went through this end to end against the base branch, wiring all checks out — every env the units set has a consumer reading it with the same legacy default, calibration writer/reader land on the same path, the sed is shell + posix correct. nice work getting it on a real pod 3. couple small things below

scripts/install
```diff
@@ -55,15 +56,51 @@ while [ $# -gt 0 ]; do
...
+else
+  _rootfs_total_mb=$(df -m / 2>/dev/null | awk 'NR==2{print $2}')
+  _rootfs_total_mb=${_rootfs_total_mb:-0}
+  if mountpoint -q /persistent 2>/dev/null; then
+    _persistent_total_mb=$(df -m /persistent 2>/dev/null | awk 'NR==2{print $2}')
+    _persistent_total_mb=${_persistent_total_mb:-0}
+  else
+    _persistent_total_mb=0
+  fi
+  if [ "$_persistent_total_mb" -gt "$_rootfs_total_mb" ]; then
```
ng, line 90: this one picks by total partition size but the node_modules + swap pickers go by free space with a headroom floor. fine for the db since it grows unbounded, but its a silent divergence from the sibling pickers — can you drop a one line comment here saying why its total not free so nobody comes along later and "fixes" it into consistency

scripts/install
```diff
@@ -617,8 +654,29 @@ fi
...
+# Migrate from a legacy /persistent/sleepypod-data layout if the picker
+# now wants a different partition. Same empty-target guard as the
+# node_modules migration in ensure_persistent_node_modules() — avoids
+# mixing DBs from two different installs.
+LEGACY_DATA_DIR="/persistent/sleepypod-data"
+if [ "$DATA_DIR" != "$LEGACY_DATA_DIR" ] && [ -d "$LEGACY_DATA_DIR" ] \
+   && [ -z "$(ls -A "$DATA_DIR" 2>/dev/null)" ]; then
+  echo "Migrating data dir from $LEGACY_DATA_DIR to $DATA_DIR..."
+  mkdir -p "$(dirname "$DATA_DIR")"
+  mv "$LEGACY_DATA_DIR" "$DATA_DIR"
```
ng, line 671: if /sleepypod-data already exists but empty (aborted prior run) this mv nests it as /sleepypod-data/sleepypod-data instead of renaming. the guard above allows an empty target so a rmdir "$DATA_DIR" 2>/dev/null || true right before the mv would close it (no-op when absent)

scripts/bin/sp-update
```diff
@@ -304,6 +310,16 @@ if [ -d "$INSTALL_DIR/modules" ]; then
   if ! command -v uv &>/dev/null; then
     echo "Warning: uv not found — biometrics modules not synced. Re-run the full installer."
   fi
+  # Probe the firmware's /usr/bin/python3 for stdlib integrity. Pod 5
```
ng, line 313: says pod 5 here but install and capabilities both say pod 3, should be pod 3

PR #645, opened 2026-06-14
Persist downsampled capacitive frames for historical replay

ng:
## Why

The Automations right pane shows the capacitive presence matrix (bed-pressure zones) only as a **live** view — the raw ~2 Hz `capSense2` frames are never persisted, so the spatial zone replay and `cap.*` backtest can't reach past nights. This is the foundation for lifting the night/date selector to pane scope and adding a standalone "Last Night" historical view (follow-up PRs).

## What

Add a downsampling writer that aggregates the live capacitive stream into **one row per side per ~5s window** in a new `cap_sense_frames` table, pruned to **~48h** (the bulkiest sensor stream — only the last night or two is ever replayed).

- New `cap_sense_frames` table (`zones` head/torso/legs means, scalar `max/mean/spread`, modal `peakZone`, `frameCount`).
- `streaming/capFramePersistence.ts`: ts-keyed windowing (correct under stream catch-up), best-effort writes (a DB failure never disrupts the live stream), throttled 48h prune.
- Wired into the `piezoStream` broadcast loop alongside the existing live snapshot; windows reset on RAW-file switch.
- Extracted `reduceCap` + new `zoneTriple` into dependency-free `automation/capReduce.ts` to avoid a `piezoStream ↔ signals.biometrics` import cycle.

Scalar (Pod 3) sensors persist `zones: null` — no spatial resolution to replay, but `cap.*` scalars still backtest.

## Scope / non-goals

This PR only **writes** the frames. Reading them into the backtest series and the UI (pane-level night picker, standalone history view) are follow-up PRs.

## Test plan

- `vitest run src/automation/tests/capReduce.test.ts src/streaming/tests/capFramePersistence.test.ts` — new unit coverage for `zoneTriple` and the window state machine (accumulate within window, rollover past the 5s boundary, scalar→null zones, per-side independence, reset).
- `vitest run src/automation/tests/signals.biometrics.test.ts src/streaming/tests/piezoStream.test.ts src/lib/tests/occupancy.test.ts` — existing suites green after the `reduceCap` extraction.
- `tsc` clean; `eslint` clean on changed files.
- `pnpm db:biometrics:generate` → migration `0010_fixed_silvermane.sql`.

Manual: on a pod, confirm `cap_sense_frames` accrues ~1 row/side/5s while occupied and that rows older than 48h are pruned.

## Summary by CodeRabbit

* **New Features**
  * Added historical storage and analysis of capacitive presence sensor readings with spatial zone tracking
  * Automatic data retention and cleanup for storage optimization

* **Refactor**
  * Reorganized internal sensor data processing utilities

<details open>

```bash
sudo sp-update feat/persist-cap-frames
```

```bash
curl -fsSL https://raw.githubusercontent.com/sleepypod/core/feat/persist-cap-frames/scripts/install \
  | sudo bash -s -- --branch feat/persist-cap-frames
```

🎯 **Pin to this exact build** ([run 27516162063](https://github.com/sleepypod/core/actions/runs/27516162063) · `34b060c`) — fresh-pod path, locked to one CI artifact for reproducible review:
```bash
curl -fsSL https://raw.githubusercontent.com/sleepypod/core/34b060c475e873eebcd6765f33607daae66029ae/scripts/install \
  | sudo bash -s -- \
      --branch feat/persist-cap-frames \
      --artifact-url 'https://nightly.link/sleepypod/core/actions/runs/27516162063/sleepypod-core.zip'
```

🧊 Artifact: [`sleepypod-core`](https://github.com/sleepypod/core/actions/runs/27516162063/artifacts/7626417573) · self-destructs in 30 days 💥
</details>

### DIFF
--- a/src/automation/capReduce.ts
+++ b/src/automation/capReduce.ts
@@ -1,0 +1,50 @@
+/**
+ * Pure reducers over a capacitive presence channel array.
+ *
+ * capSense2 carries `[A1,A2,B1,B2,C1,C2,ref1,ref2]` per side — body-contact load
+ * across head/torso/legs (NOT temperature) plus two reference channels. capSense
+ * (Pod 3) carries a single scalar. These helpers collapse either shape to scalar
+ * stats (for the automation engine) or a three-zone triple (for the UI/replay).
+ *
+ * Kept dependency-free so both the engine (`signals.biometrics`) and the
+ * streaming persistence writer (`streaming/capFramePersistence`) can import it
+ * without pulling in a module cycle.
+ */
+
+export function mean(xs: number[]): number {
+  return xs.reduce((a, b) => a + b, 0) / xs.length
+}
+
+/**
+ * Reduce a per-side capacitive channel array to scalar matrix signals. Returns
+ * null when there are no usable channels. `peakZone` is the index (0–2) of the
+ * highest paired zone (A/B/C), available only on the full 6-channel frame.
+ */
+export function reduceCap(values: number[]): { max: number, mean: number, spread: number, peakZone: number | null } | null {
+  // capSense2 carries 6 sensor channels + 2 reference channels; drop the refs.
+  const zones = values.length >= 8 ? values.slice(0, 6) : values
+  if (zones.length === 0) return null
+  const max = Math.max(...zones)
+  const min = Math.min(...zones)
+  // Three paired zones (A/B/C) only when the full 6-channel frame is present.
+  const peakZone = zones.length === 6
+    ? [mean([zones[0], zones[1]]), mean([zones[2], zones[3]]), mean([zones[4], zones[5]])]
+        .reduce((best, v, i, arr) => (v > arr[best] ? i : best), 0)
+    : null
+  return { max, mean: mean(zones), spread: max - min, peakZone }
+}
+
+/**
+ * Collapse a channel array to the three paired-zone (head/torso/legs) means used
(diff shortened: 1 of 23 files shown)


ng:
@coderabbitai review

ng:
@coderabbitai review

ng:
Thanks @coderabbitai — all four addressed in the latest push:

1. **`peakZone` drift (`zoneTriple` gating)** — tightened `zoneTriple` to `zones.length !== 6` so it returns non-null *iff* `reduceCap` also emits a `peakZone`. A frame can no longer accumulate zones without a matching peak vote.
2. **One-row-per-window** — `idx_cap_sense_frames_side_ts` → `uniqueIndex('uq_cap_sense_frames_side_ts')`, regenerated migration `0010`. The writer now uses `.onConflictDoNothing()` so a restart re-reading the active RAW file dedups silently instead of throwing per row.
3. **No persistence when idle (critical)** — the file-tailing loop's `wss.clients.size === 0` early return was dropping all overnight history. Changed the guard to `if (!wss) return` so downsampling continues with zero clients; broadcast stays per-client guarded, so an idle loop does no fan-out.
4. **Non-hermetic test** — mocked `@/src/db` in the persistence suite so the rollover case no longer hits the unmigrated `cap_sense_frames` table; the error/throttle cases re-spy per test.

`tsc`, `eslint`, and the affected vitest suites are green.

PR #669, opened 2026-07-21
fix(server): acknowledge restart-orphaned pump alerts by falling back to the newest active row

ng:
## Problem

Field report (eight-pod bundle, 2026-07-21): pump alerts 38/39 (`stall_left`/`stall_right`, `ack=null`) became permanently unacknowledgeable after an sp-update restarted the server. The guard's `activeAlertId` lives on `globalThis`, so `acknowledgeAndRestore` got `alertId=null` back, stamped nothing, and the rows sat in the `/api/pump-alerts` active list forever. The only remaining cleanup path was per-row `dismissAlert` by id from the diagnostics console.

## Fix

Root cause first: the guard now **rehydrates at startup**. For each side, the newest un-acknowledged, un-dismissed `power_off` row re-arms the in-memory block, restore snapshot, and banner notice (skipped when `pumpStallProtectionEnabled` is off), so after a restart the banner reappears and acknowledge/dismiss work through the normal in-memory path. Wired in `instrumentation.ts` before keepalives start; DB errors warn instead of crashing startup.

The acknowledge path was then hardened around it, following the adversarial-review findings:

- The orphan fallback now also selects the row's persisted `restoreTargetTemperature`/`restoreDurationSeconds` (ADR 0022) and replays them through the device router — but only when a status read shows the side still parked; a failed status read skips the replay conservatively. A failed orphan lookup now surfaces as `INTERNAL_SERVER_ERROR` instead of a silent success, the lookup orders by `desc(timestamp), desc(id)` for same-second determinism, and the response gained `orphanRecovered` so clients can tell a restart recovery from a plain no-op.
- `acknowledgedAt` is stamped only **after** both restore calls succeed. On restore failure the row stays active and the guard re-arms (new `pumpStallGuard.rearm`) with the original trip metadata, so the banner returns; a partial failure (power-on succeeded, setpoint failed) parks the side again before re-arming.
- `acknowledgeAndRestore` and `dismissNotification` accept an optional `alertId` (OpenAPI backward compatible); the banner passes the notice's id through both mutations, so a stale or replayed request stamps exactly the incident the user saw — including across restarts — instead of a newest-row guess. All stamp UPDATEs re-check `isNull(acknowledgedAt)`/`isNull(dismissedAt)` in the WHERE clause, mirroring `dismissAlert`'s guard against a concurrent dismiss.
- `trip()` no longer swallows a failed hardware power-off: cutoff-pending is tracked per side and retried on every subsequent frame until the command is confirmed sent.
- Banner: both actions disable while either mutation is pending, the container is `role="alert"`, the X button has an accessible name, and the dismiss comment now says what dismiss actually does (clears the notification and re-arms stall protection; the side stays off until the user powers it back on).

## Test plan

- `vitest run src/server/routers/tests/pumpAlerts.test.ts` — 41 pass: orphan replay (parked / already-powered / status-read-failure), stamp-after-restore ordering, re-arm on failure including the partial-restore park, client-id correlation for ack and dismiss, TOCTOU where-clauses, `desc(timestamp), desc(id)` ordering, lookup failure → 500. The DB mock now throws on any un-enqueued call.
- `vitest run src/hardware/tests/pumpStallGuard.test.ts` — 40 pass: rehydration (restores state+notice, skips when protection disabled, ignores acknowledged/dismissed/non-power_off rows, same-second id tiebreak, tolerates read failure), `rearm`, cutoff retry.
- New `PumpStallNotification.test.tsx` (6 pass) pins alert-id pass-through, dual-pending lockout, and the `role="alert"` contract; instrumentation tests pin rehydration running before keepalives and boot surviving a rehydration throw.
- Full suite: 3227 pass / 1 skipped. `pnpm tsc` and `pnpm lint` clean.
- Manual: on a pod with a stranded active alert, restart the server — the banner reappears; Re-enable restores the pre-stall setpoint and the row leaves the active list.

Refs: ygg `sleepypod-core-8`, PR #663.

<details open>

```bash
sudo sp-update fix/pump-alert-ack-restart
```

```bash
curl -fsSL https://raw.githubusercontent.com/sleepypod/core/fix/pump-alert-ack-restart/scripts/install \
  | sudo bash -s -- --branch fix/pump-alert-ack-restart
```

🎯 **Pin to this exact build** ([run 29812820241](https://github.com/sleepypod/core/actions/runs/29812820241) · `0649261`) — fresh-pod path, locked to one CI artifact for reproducible review:
```bash
curl -fsSL https://raw.githubusercontent.com/sleepypod/core/064926121a902b987431a8b12e371885609aee8f/scripts/install \
  | sudo bash -s -- \
      --branch fix/pump-alert-ack-restart \
      --artifact-url 'https://nightly.link/sleepypod/core/actions/runs/29812820241/sleepypod-core.zip'
```

🧊 Artifact: [`sleepypod-core`](https://github.com/sleepypod/core/actions/runs/29812820241/artifacts/8488141864) · self-destructs in 30 days 💥
</details>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved alert acknowledgment after a guard restart when the in-memory alert reference is lost.
  - When appropriate, the app now locates and acknowledges the newest matching unresolved power-off alert for the affected side.
  - Recovery succeeds even if no matching alert is found or if the lookup fails, without blocking the operation.
- **Tests**
  - Expanded automated coverage for restart and orphaned-alert acknowledgment scenarios, including error-tolerant behavior.

### DIFF
--- a/src/server/routers/pumpAlerts.ts
+++ b/src/server/routers/pumpAlerts.ts
@@ -122,9 +122,12 @@ export const pumpAlertsRouter = router({
       // A restart wipes the guard's in-memory state, so a trip from before
       // the restart has no activeAlertId anymore and its row would strand
       // unacknowledged in the active list forever. Fall back to the newest
-      // active power_off row for this side.
+      // active power_off row for this side — but only when the restore
+      // snapshot is gone too: a failed alert INSERT at trip time also
+      // returns a null id while the snapshot survives, and falling back
+      // there would stamp an older, unrelated row.
       let stampId = alertId
-      if (stampId == null) {
+      if (stampId == null && restore == null) {
         try {
           const [orphan] = await biometricsDb
             .select({ id: pumpAlerts.id })
--- a/src/server/routers/pumpAlerts.ts
+++ b/src/server/routers/pumpAlerts.ts
@@ -110,7 +110,7 @@ export const pumpAlertsRouter = router({
    */
   acknowledgeAndRestore: publicProcedure
     .meta({ openapi: { method: 'POST', path: '/pump-alerts/acknowledge', protect: false, tags: ['Pump Alerts'] } })
-    .input(z.object({ side: sideSchema }).strict())
+    .input(z.object({ side: sideSchema, alertId: idSchema.optional() }).strict())
     .output(z.object({
       success: z.boolean(),
       restoredTarget: z.number().nullable(),
@@ -123,15 +123,20 @@ export const pumpAlertsRouter = router({
       const priorNotice = getPumpStallNotice(input.side)
       const { restore, alertId } = guardAcknowledge(input.side)
 
-      // A restart wipes the guard's in-memory state, so a trip from before
-      // the restart has no activeAlertId anymore and its row would strand
-      // unacknowledged in the active list forever. Fall back to the newest
-      // active power_off row for this side — but only when the restore
-      // snapshot is gone too: a failed alert INSERT at trip time also
-      // returns a null id while the snapshot survives, and falling back
-      // there would stamp an older, unrelated row. The row's persisted
-      // restore columns (ADR 0022) stand in for the lost snapshot.
-      let stampId = alertId
+      // Identity correlation: a client that saw the banner passes the
+      // notice's alertId, so a stale or replayed request stamps exactly
+      // that incident's row (and never falls through to guessing).
+      //
+      // Without a client id, a restart wipes the guard's in-memory state,
+      // so a trip from before the restart has no activeAlertId anymore and
+      // its row would strand unacknowledged in the active list forever.
+      // Fall back to the newest active power_off row for this side — but
+      // only when the restore snapshot is gone too: a failed alert INSERT
+      // at trip time also returns a null id while the snapshot survives,
+      // and falling back there would stamp an older, unrelated row. The
+      // row's persisted restore columns (ADR 0022) stand in for the lost
+      // snapshot.
+      let stampId = input.alertId ?? alertId
       let orphanRecovered = false
       let orphanRestore: { targetTemperature: number, durationSeconds: number } | null = null
       if (stampId == null && restore == null) {
@@ -175,14 +180,25 @@ export const pumpAlertsRouter = router({
       }
 
       // Best-effort by design: a stamp failure only leaves the row in the
-      // active list; the guard itself is already released.
+      // active list; the guard itself is already released. The WHERE clause
+      // re-checks activity so a concurrently dismissed row is not stamped
+      // twice (mirroring dismissAlert), and a client-provided id must also
+      // prove it names this side's power_off incident.
       const stampAcknowledged = async (): Promise<void> => {
         if (stampId == null) return
+        const conditions = [
+          eq(pumpAlerts.id, stampId),
+          isNull(pumpAlerts.acknowledgedAt),
+          isNull(pumpAlerts.dismissedAt),
+        ]
+        if (input.alertId != null) {
+          conditions.push(eq(pumpAlerts.side, input.side), eq(pumpAlerts.action, 'power_off'))
+        }
         try {
           await biometricsDb
             .update(pumpAlerts)
             .set({ acknowledgedAt: new Date() })
-            .where(eq(pumpAlerts.id, stampId))
+            .where(and(...conditions))
         }
         catch (err) {
           console.warn('[pumpAlerts] failed to stamp acknowledgedAt:', err instanceof Error ? err.message : err)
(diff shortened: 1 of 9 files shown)


src/server/routers/pumpAlerts.ts
```diff
@@ -119,12 +119,40 @@ export const pumpAlertsRouter = router({
...
-      if (alertId != null) {
+      // A restart wipes the guard's in-memory state, so a trip from before
+      // the restart has no activeAlertId anymore and its row would strand
+      // unacknowledged in the active list forever. Fall back to the newest
+      // active power_off row for this side — but only when the restore
+      // snapshot is gone too: a failed alert INSERT at trip time also
+      // returns a null id while the snapshot survives, and falling back
+      // there would stamp an older, unrelated row.
+      let stampId = alertId
+      if (stampId == null && restore == null) {
```
ng, line 142: **🟡 Restart heuristic could stamp an unrelated row** (Correctness) — **fixed in f76699c**

`trip()` also returns a null alert id when its `pump_alerts` INSERT fails, while the in-memory restore snapshot survives. Treating any null id as a restart let the fallback acknowledge an older, unrelated `power_off` row for the side. The fallback is now gated on `restore == null` as well (a real restart wipes both), with a pinning test for the failed-insert case.

<details>
<summary>Review context</summary>

- Source: Optimizer (sonnet + codex, cross-vendor)
- Skeptic verdicts: sonnet ✅ 85, codex 🔄 95
- Impact: audit-trail integrity only — the device restore path already used the correct in-memory snapshot
</details>

src/server/routers/pumpAlerts.ts
```diff
@@ -119,12 +119,40 @@ export const pumpAlertsRouter = router({
...
+        try {
+          const [orphan] = await biometricsDb
+            .select({ id: pumpAlerts.id })
```
ng: **🟡 Persisted restore snapshot is not replayed** (Correctness) — needs author decision, not auto-fixed

The row's `restoreTargetTemperature`/`restoreDurationSeconds` columns exist so "an acknowledgement can restore it through the normal command path" (schema comment, ADR 0022 §Control intervention), but the fallback selects only `id` — after a restart, Re-enable stamps the row and returns `restoredTarget: null`, leaving the side off with no feedback. No code path reads these columns back.

Both reviewers confirmed the gap (95/100 confidence); the remedy is disputed: replaying the snapshot blindly could re-energize a side that was already restored after a silent stamp failure, so a safe version needs a parked-state check. The PR body frames no-restore as an intentional conservative choice — if that stands, consider documenting it in ADR 0022; if not, select the two columns and restore behind a parked-state guard.

<details>
<summary>Review context</summary>

- Source: Optimizer (codex), ADR/schema claims verified by review lead
- Skeptic verdicts: sonnet ✅ 95, codex 🔄 100
- Status: disputed remedy — author decision
</details>

src/server/routers/pumpAlerts.ts
```diff
@@ -119,12 +119,40 @@ export const pumpAlertsRouter = router({
...
+        }
+      }
+
+      if (stampId != null) {
         try {
           await biometricsDb
             .update(pumpAlerts)
             .set({ acknowledgedAt: new Date() })
-            .where(eq(pumpAlerts.id, alertId))
+            .where(eq(pumpAlerts.id, stampId))
```
ng: **🟢 Fallback UPDATE lacks an active-row re-check** (Race condition) — suggestion, not applied

The fallback SELECTs the newest active row, then UPDATEs by `id` alone. A concurrent `dismissAlert` on that row between the two lands the row with both `dismissedAt` and `acknowledgedAt` set. `dismissAlert` guards its own update with `isNull(dismissedAt)` — mirroring that here (`and(eq(id, stampId), isNull(dismissedAt), isNull(acknowledgedAt))`) closes the window.

<details>
<summary>Review context</summary>

- Source: Skeptic (sonnet), new finding; Haiku isolation score 85
- Severity: Minor — narrow window, benign-ish outcome
</details>

ng (commented):
## Adversarial review — PR #669

**Depth**: standard + Codex cross-vendor sidecar (escalation score 3: baseline +1, modified api endpoint +2). Change types: api, test. Reviewers: Sonnet Optimizer/Skeptic (Claude), Codex Optimizer/Skeptic (sidecar), Haiku isolation scorers. Mechanical: lint/tsc clean, 3198 tests green.

**Fixed in f76699c**
- Gated the orphan fallback on `restore == null` so a failed alert INSERT at trip time can no longer stamp an unrelated older row (cross-vendor consensus)
- Pinned the fallback query contract in tests: `desc(pumpAlerts.timestamp)`, `limit(1)`, exact `eq`/`isNull` columns, update id — mutation-testing hardening

**Needs author decision (posted inline)**
- Persisted restore snapshot (`restoreTargetTemperature`/`restoreDurationSeconds`) is never replayed after a restart — confirmed against ADR 0022 intent by both vendors; remedy disputed (blind replay is unsafe after silent stamp failures)

**Noted, not fixed** (minor/pre-existing, candidates for follow-up issues)
- Fallback UPDATE lacks `isNull(dismissedAt)` re-check (inline)
- Lookup failure returns `success: true` (matches file convention; PR states intentional)
- Mutation is non-idempotent with multiple active rows — proper fix needs a client-supplied alert id
- `dismissNotification` has the same restart-orphan gap (CodeRabbit also flagged) — but a side-only fallback there would inherit the same defects; needs identity correlation
- Pre-existing class issues confirmed by both Skeptics: restart drops the guard interlock + banner entirely; WS status frames omit `pumpStallNotifications` (banner invisible while WS healthy); scheduler/Autopilot/HomeKit/gesture paths energize sides without consulting `shouldBlock`; ack stamps before hardware restore succeeds; failed cutoff in `trip()` is swallowed

**Verdict**: no regression found in the diff itself (both Skeptics agree); Codex rates the restart story incomplete end-to-end (banner unreachable post-restart), which is pre-existing scope. Recommendation: **Comment** — mergeable as a narrow fix; the restart-rehydration class issue deserves its own ticket.

PR #670, opened 2026-07-21
fix(hardware): pump-stall guard hardening — HomeKit gate, twitchy-trip, active recovery, scheduled reboot

ng:
> **This PR bundles four pump-stall guard changes.** The original HomeKit gate (§1), plus three fixes from a 2026-07-29 field report: the daily reboot never ran, the guard tripped on ~2s glitches, and auto-recovery could never fire overnight.

---

## 1. HomeKit gate (original)

Field report (eight-pod bundle, 2026-07-21): with both sides guard-blocked, the web UI correctly refused power/temp commands (PRECONDITION_FAILED at `device.ts:325/437`), but adjusting the thermostat in the Home app silently re-powered the side — `sideController` wrote straight to the shared hardware client with no `shouldBlock` consultation. With a genuine stall this bypass lets a user keep re-heating a side with zero flow, defeating the fail-safe.

`setSidePowerOn` and `setTargetTemperature`'s firmware-push path now refuse when `pumpStallGuard.shouldBlock(side)` is true, matching the device router and keepalive. `setSidePowerOn` refuses before touching the intent latch (onGet keeps reporting the true off state); `setTargetTemperature` still caches the requested target (staging on an off side stays allowed) but gates the hardware push, covering the stuck-latch case where `reconcileIntendedPower` never clears a latched ON after a trip; `setSidePowerOff` is never gated (safe direction, used by the guard's own trip path). The thrown error rejects the HAP onSet so Home surfaces the failed write. A follow-up commit surfaces the guard-blocked staged target on the alert card.

## 2. Scheduled reboot runs as the `sleepypod` user

The daily / pre-prime reboot jobs shelled `systemctl reboot`, but next-server runs as `User=sleepypod`, which can't reboot directly — the command failed interactive-auth and the pod never rebooted on schedule (observed: 5-day uptime with `reboot_daily` on; `pre-prime-calibration` and `daily-prime` ran fine). `executeReboot` now shells `sudo -n systemctl reboot` via a NOPASSWD rule the installer adds to `/etc/sudoers.d/sleepypod-update`, scoped to exactly `systemctl reboot` and mirroring `system.triggerUpdate`.

`sp-maintenance` re-applies the sudoers rule (idempotently, best-effort) on every service start, and `sp-update` installs the new `sp-maintenance` before it restarts the service — so **`sp-update <branch>` is enough** for already-installed pods; no full reinstall needed.

## 3. Twitchy trip hardening

The guard powered a side off (and, without recovery, kept it off all night) on ~2 seconds of bad readings — e.g. a single frame where **both** sides read exactly 0 RPM while flow/loop-temp kept moving, both healthy again a minute later. Two independent fixes:

- **Bilateral-zero de-glitch** (`deviceStateSync`): both pumps dropping from clearly running to exactly 0 RPM in one frame is the dropped/garbled-frame signature, not a real dual stall — which hits one side. That single frame is suppressed so neither dwell counter advances; a sustained dual-stall still trips one frame later.
- **Time-based dwell floor** (`pumpStallGuard`): a trip now requires the pump to stay sub-threshold for both `dwellSamples` frames **and** `DWELL_MIN_MS` (10s) of wall clock, so a brief low burst can't end a night. Frame time is threaded through `onFrame` so the floor is driven by frame arrival and stays deterministic under test.

## 4. Active-probe auto-recovery

`pumpStallAutoRecoveryEnabled` was inert for the case that matters. A trip powers the pump off, so RPM sits at 0 forever and the "recoverySamples consecutive healthy frames" watcher never advances — nothing re-energizes the pump to let it prove itself. Daytime trips only recovered because the daily prime or a manual restart happened to spin the pump up; overnight trips stayed `power_off` until morning (field trace: trip 02:16 → 0 RPM for 11.7 h → 14:00 prime).

The guard now **actively probes**: a blocked side with auto-recovery enabled is re-energized after a backoff (5 m → 15 m → 30 m); sustained healthy RPM restores it via the existing `autoRecover` path, otherwise the probe is powered back off and the next backoff applies, staying off after the last attempt until acknowledged. Probe energize / de-energize serialize through `withSideLock`. ADR 0022 gains a Revisions section for §3–§4.

## Test plan

- **Full unit suite: 3330 pass / 0 fail.** `tsc` and `eslint` clean.
- `vitest run src/hardware/tests/pumpStallGuard.test.ts` — 59 pass, incl. 4 time-dwell-floor cases and 6 active-probe cases (backoff timing, restore on spin-up, power-off + give-up after 3 attempts, no-snapshot, energize-failure, side-lock serialization).
- `vitest run src/hardware/tests/deviceStateSync.stallSuppression.test.ts` — 4 new bilateral-zero cases (drop after both-running, feed sustained zero, single-side not suppressed, cold-start not suppressed).
- `vitest run src/scheduler/tests/` — reboot command pinned to `sudo -n systemctl reboot`.
- Manual (§1): trip the guard and try powering the side on from Home — write fails, pump stays off; acknowledge, control returns.

Refs: ygg `sleepypod-core-7`, PR #663; field report Discord 2026-07-29.

<details open>

```bash
sudo sp-update fix/homekit-stall-guard-gate
```

```bash
curl -fsSL https://raw.githubusercontent.com/sleepypod/core/fix/homekit-stall-guard-gate/scripts/install \
  | sudo bash -s -- --branch fix/homekit-stall-guard-gate
```

🎯 **Pin to this exact build** ([run 29962457341](https://github.com/sleepypod/core/actions/runs/29962457341) · `c20e7ec`) — fresh-pod path, locked to one CI artifact for reproducible review:
```bash
curl -fsSL https://raw.githubusercontent.com/sleepypod/core/c20e7ec520996395150cbf69c571f0e6048afb70/scripts/install \
  | sudo bash -s -- \
      --branch fix/homekit-stall-guard-gate \
      --artifact-url 'https://nightly.link/sleepypod/core/actions/runs/29962457341/sleepypod-core.zip'
```

🧊 Artifact: [`sleepypod-core`](https://github.com/sleepypod/core/actions/runs/29962457341/artifacts/8546532306) · self-destructs in 30 days 💥
</details>

### DIFF
--- a/src/homekit/accessories/sideController.ts
+++ b/src/homekit/accessories/sideController.ts
@@ -124,7 +124,19 @@ export function isEffectivelyPowered(monitor: DacMonitor, side: Side): boolean {
 export function reconcileIntendedPower(status: DeviceStatus, side: Side): void {
   const intended = intendedPower[side]
   if (intended === null) return
-  if (isPoweredFromStatus(status, side) === intended) {
+  const observed = isPoweredFromStatus(status, side)
+  if (observed === intended) {
+    intendedPower[side] = null
+    return
+  }
+  // A stall trip can land after a power-on was accepted but before any
+  // status frame confirms it: firmware keeps reporting OFF, so the ON latch
+  // would survive forever ("write in flight" never resolves). Once the guard
+  // clears, a mere slider drag would then compute powered=true from the
+  // stale latch and re-energize the side. Drop the latch only when the
+  // guard blocks the side AND firmware observes OFF — a normal in-flight
+  // ON (guard healthy) must keep its latch.
+  if (intended && !observed && pumpStallShouldBlock(side)) {
     intendedPower[side] = null
   }
 }
--- a/src/homekit/accessories/sideController.ts
+++ b/src/homekit/accessories/sideController.ts
@@ -17,6 +17,8 @@
  * across off-cycles.
  */
 
+import { HapStatusError } from 'hap-nodejs'
+import type { HAPStatus } from 'hap-nodejs'
 import type { DacMonitor } from '@/src/hardware/dacMonitor'
 import type { DeviceStatus, Side } from '@/src/hardware/types'
 import { MAX_TEMP, MIN_TEMP, TEMP_NEUTRAL } from '@/src/hardware/types'
@@ -24,6 +26,25 @@ import { getSharedHardwareClient } from '@/src/hardware/dacMonitor.instance'
 import { getAutomationEngineIfRunning } from '@/src/automation'
 import { shouldBlock as pumpStallShouldBlock } from '@/src/hardware/pumpStallGuard'
 import { withSideLock } from '@/src/hardware/sideLock'
+
+/**
+ * Guard denials surface as NOT_ALLOWED_IN_CURRENT_STATE (-70412): hap-nodejs
+ * maps any plain Error to SERVICE_COMMUNICATION_FAILURE (-70402), which iOS
+ * renders as a bridge outage rather than an intentional refusal. Message
+ * wording mirrors the REST path (server/helpers.ts) — the UI action is
+ * literally "Re-enable".
+ */
+// HAPStatus is an ambient const enum in hap-nodejs typings — inaccessible as
+// a value under isolatedModules, so pin the wire constant directly (asserted
+// numerically in sideController.test.ts).
+const NOT_ALLOWED_IN_CURRENT_STATE = -70412 as HAPStatus
+
+class GuardBlockedError extends HapStatusError {
+  constructor() {
+    super(NOT_ALLOWED_IN_CURRENT_STATE)
+    this.message = 'Pump stall protection active — re-enable the side first'
+  }
+}
 
 /**
  * HomeKit must honor the pump stall guard like every other write surface
@@ -35,7 +56,7 @@ import { withSideLock } from '@/src/hardware/sideLock'
 function assertNotGuardBlocked(side: Side, label: string): void {
   if (!pumpStallShouldBlock(side)) return
   console.warn(`[homekit] refused ${label} — pump stall protection active on ${side}`)
-  throw new Error('Pump stall protection active — acknowledge the alert first')
+  throw new GuardBlockedError()
 }
 
 const lastTargetF: Record<Side, number | null> = { left: null, right: null }
@@ -169,12 +190,10 @@ export async function setSidePowerOn(monitor: DacMonitor, side: Side): Promise<v
   intendedPower[side] = true
   try {
     await withSideLock(side, async () => {
-      // Throw rather than silently skip: the catch below rolls intendedPower
-      // back so iOS Home reverts the toggle instead of showing a phantom ON.
-      if (pumpStallShouldBlock(side)) {
-        console.warn(`[homekit] skipped setPower(${side}, true): pump stall guard blocks ${side}`)
-        throw new Error(`pump stall protection active on ${side}`)
-      }
+      // Re-check inside the lock: a trip can land while this call is queued.
+      // Throw rather than silently skip: the catch below resets intendedPower
+      // so iOS Home reverts the toggle instead of showing a phantom ON.
+      assertNotGuardBlocked(side, `setPower(${side}, true)`)
       const target = clampF(getStagedTargetF(monitor, side))
       lastTargetF[side] = target
       registerManualOverride(side)
--- a/src/homekit/accessories/sideController.ts
+++ b/src/homekit/accessories/sideController.ts
@@ -42,6 +42,10 @@ const NOT_ALLOWED_IN_CURRENT_STATE = -70412 as HAPStatus
 class GuardBlockedError extends HapStatusError {
   constructor() {
     super(NOT_ALLOWED_IN_CURRENT_STATE)
+    // The parent constructor pins the prototype to HapStatusError.prototype
+    // (its Error-subclass workaround), which breaks instanceof for further
+    // subclasses — restore ours so the rollback catch can identify us.
+    Object.setPrototypeOf(this, GuardBlockedError.prototype)
     this.message = 'Pump stall protection active — re-enable the side first'
   }
 }
@@ -204,7 +208,12 @@ export async function setSidePowerOn(monitor: DacMonitor, side: Side): Promise<v
     })
   }
   catch (e) {
-    intendedPower[side] = prev
+    // A guard rejection means the side is definitively de-energized (the
+    // trip path forces power off), so latch false rather than restoring
+    // `prev`: with two blocked power-ons in flight, `prev` can be the other
+    // call's transient ON, and restoring it would resurrect a phantom ON
+    // with zero hardware writes.
+    intendedPower[side] = e instanceof GuardBlockedError ? false : prev
     throw e
   }
 }
--- a/src/homekit/accessories/sideController.ts
+++ b/src/homekit/accessories/sideController.ts
@@ -165,6 +165,16 @@ export function getStagedTargetF(monitor: DacMonitor, side: Side): number {
  *   matches iOS Home thermostat-tile semantics. Power resumes through
  *   TargetHeatingCoolingState or the dedicated Power switch.
  *
+ * Requested-intent contract (PR #670 review, F1 — decided 2026-07-22): the
+ * cache retains the requested value even when the pump-stall gate rejects
+ * the firmware push. iOS Home shows the slider reverting (HAP keeps the
+ * characteristic at oldValue on rejection) while the staged value survives
+ * and applies on the next explicit power-on. This is deliberate: lastTargetF
+ * records what the user asked for, not what firmware confirmed, and heating
+ * to it requires a separate deliberate power-on after the side is re-enabled.
+ * The web UI's pump-stall alert card — the only surface that can re-enable
+ * the side — is where the staged/visible mismatch gets surfaced.
+ *
  * f is captured in closure so back-to-back drags of the slider each push
  * their own value, even though the cache only retains the latest.
  */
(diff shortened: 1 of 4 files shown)


src/homekit/accessories/sideController.ts
```diff
@@ -129,12 +142,10 @@ export async function setTargetTemperature(
...
-    if (pumpStallShouldBlock(side)) {
-      // Keep the staged target (same semantics as adjusting while OFF) but
-      // never push a write that would re-energize a parked side.
-      console.warn(`[homekit] skipped setTemperature(${side}): pump stall guard blocks ${side}`)
-      return
-    }
+    // Gate only the firmware push: staging a target on an off side is
+    // harmless, and the guard trip leaves the intent latch stuck ON
+    // (firmware never confirms), so `powered` alone can't be trusted here.
+    assertNotGuardBlocked(side, `setTemperature(${side}, ${f})`)
```
ng, line 195: **🟡 Rejected temperature persists in the hidden power-on cache** (Correctness) — **DISPUTED, author decision needed**

`lastTargetF[side] = f` runs before this gate. When the guard blocks a powered-side push, the throw makes HAP freeze the visible characteristic at its old value (hap-nodejs `handleSetRequest` only assigns `this.value` on the success branch), but the hidden cache keeps the rejected setpoint — a later power-on via `getStagedTargetF` heats to a value Home reported as failed.

The reviewers split cross-vendor:
- **Claude Skeptic (agree, 90)**: new regression — pre-PR, blocked writes silently succeeded so cache and visible value agreed; fix by not persisting (or rolling back) `lastTargetF` on the gated-rejection path.
- **Codex Skeptic (disagree, 95)**: `lastTargetF` is deliberately *requested intent*, not confirmed HAP state; the retention-after-rejection contract is pinned by this PR's own test, and heating still requires an explicit power-on. Both proposed fixes regress something (lose staged intent, or hide the safety denial).

**Options:** (a) keep throw + roll back the cache on guard rejection; (b) revert to dev's warn-and-return for blocked temp pushes (stage silently, throw only for power-on); (c) keep as-is and document requested-intent semantics.

<details><summary>Review context</summary>

- Source: Codex Optimizer (confidence 98)
- Skeptic verdict: split — Claude ✅ 90 / Codex ⚠️ 95
- Model consensus: cross-vendor disagreement
- Confidence: Disputed
</details>

ng (commented):
## Adversarial review — standard depth + Codex cross-vendor sidecar

**Recommendation: Comment** — the safety gate itself is sound (both vendors agree the guard-blocked power-on/temperature refusal logic is correct and the dev merge is coherent; tsc/eslint/3279 tests green), but one 🟡 Major item is **cross-vendor disputed and needs an author decision** (rejected-setpoint cache retention, inline below), and several Minor test-strength items are worth picking up.

- Reviewers: Sonnet Optimizer + Codex Optimizer → merged → Sonnet Skeptic + Codex Skeptic → Haiku isolation scoring (7 scorers, no downgrades)
- Findings: 13 merged (1 Major disputed, 7 Minor, 2 Nit, 3 Pre-existing) + 1 Skeptic-discovered pre-existing race
- Auto-fix applied: none (the sole Major is disputed; the rest fall below the auto-fix bar)
- Pre-existing (not this PR's fault): stale ON intent latch after trip+clear (probe-confirmed), trip() cutoff outside withSideLock (tracked: ygg sleepypod-core-10), autoRecover outside withSideLock (untracked), concurrent blocked power-on rollback race (probe-confirmed)

Full artifacts in `.reviews/fix-homekit-stall-guard-gate/` (gitignored).

src/homekit/accessories/sideController.ts
```diff
@@ -153,6 +164,7 @@ export async function setTargetTemperature(
  * that the failed power-on left in an unknown state.
  */
 export async function setSidePowerOn(monitor: DacMonitor, side: Side): Promise<void> {
+  assertNotGuardBlocked(side, `setPower(${side}, true)`)
```
ng, line 214: **🟢 In-lock re-check duplicates assertNotGuardBlocked with divergent wording** (Pattern)

This ingress assert plus dev's inline re-check inside `withSideLock` (lines 174–177) are two hand-written copies of the same guard refusal with different warn text and thrown messages — identical refusals produce timing-dependent diagnostics. `device.ts` and `runOnce.ts` reuse one shared `assertPumpStallNotBlocked` for both the pre-lock and in-lock checks.

**Suggested fix:** replace the inline re-check with `assertNotGuardBlocked(side, ...)` and update the queued-trip test's expected message.

<details><summary>Review context</summary>

- Source: both Optimizers (cross-vendor consensus)
- Skeptic verdict: Agree (Claude 90, Codex 95)
- Haiku isolation score: 80
- Confidence: Very high
</details>

src/homekit/accessories/sideController.ts
```diff
@@ -25,6 +25,19 @@ import { getAutomationEngineIfRunning } from '@/src/automation'
...
 import { withSideLock } from '@/src/hardware/sideLock'
 
+/**
+ * HomeKit must honor the pump stall guard like every other write surface
+ * (device router, keepalive). Without this gate a guard-blocked side could
+ * be silently re-powered from the Home app — re-heating a side with zero
+ * flow and defeating the fail-safe. Powering OFF is never gated: it is the
+ * safe direction and the guard's own trip path uses it.
+ */
+function assertNotGuardBlocked(side: Side, label: string): void {
```
ng, line 60: **🟢 Guard denial surfaces in HAP as SERVICE_COMMUNICATION_FAILURE** (Correctness — follow-up, not PR-blocking)

A plain `Error` from an onSet handler maps to `-70402 SERVICE_COMMUNICATION_FAILURE` (probe-confirmed), so an intentional safety denial reads as an accessory/transport fault. `HapStatusError(HAPStatus.NOT_ALLOWED_IN_CURRENT_STATE)` (`-70412`) propagates correctly. Every existing accessory (primeSwitch etc.) throws plain `Error` today, so this should be a repo-wide onSet error-semantics follow-up, done after consolidating the duplicated guard checks; also align the message wording with `server/helpers.ts` ("re-enable the side first" — the UI action is literally *Re-enable*).

<details><summary>Review context</summary>

- Source: Codex Optimizer (+ Sonnet Optimizer on wording)
- Skeptic verdict: Agree with modifications (Claude 90, Codex 100)
- Haiku isolation score: 90
- Confidence: High (scoped as follow-up)
</details>

src/homekit/tests/sideController.test.ts
```diff
@@ -408,24 +409,24 @@ describe('sideController', () => {
...
       warn.mockRestore()
     })
 
-    it('rejects setSidePowerOn and rolls back intent while the guard blocks the side', async () => {
+    it('setSidePowerOn re-checks inside the lock and rolls back intent when a trip lands while queued', async () => {
       const warn = vi.spyOn(console, 'warn').mockImplementation(() => {})
-      pumpStallShouldBlock.mockReturnValue(true)
+      // Guard is clear at the ingress assert, then trips before the locked
+      // section runs — the in-lock re-check must still refuse the write.
+      shouldBlock.mockReturnValueOnce(false).mockReturnValue(true)
```
ng: **🟢 Queued-trip tests prove mock sequencing, not lock contention** (Testing)

`mockReturnValueOnce(false).mockReturnValue(true)` only proves `shouldBlock` is consulted twice — the test still passes if the second check moves *before* `withSideLock`, so it does not pin the in-lock re-check invariant its title claims. The blocked-powered-side temperature test has the same gap (starts blocked; never exercises lock-then-trip ordering), and the latch-untouched test only observes state after rejection resolves.

**Suggested fix:** use the held-lock pattern from the gesture tests (fdacebf): hold the side lock, start the call while the guard is clear, trip the guard, release in `finally`, assert rejection + no hardware call (and for the latch test, assert effective power stays OFF while pending).

<details><summary>Review context</summary>

- Source: Codex Optimizer (findings 6–8)
- Skeptic verdict: Agree (Claude 80–85, Codex 85–90)
- Haiku isolation scores: 75 / 72 / 75
- Confidence: High
</details>

src/homekit/tests/sideController.test.ts
```diff
@@ -436,12 +437,52 @@ describe('sideController', () => {
       warn.mockRestore()
     })
 
-    it('still allows setSidePowerOff while the guard blocks the side', async () => {
-      pumpStallShouldBlock.mockReturnValue(true)
+    it('setTargetTemperature refuses the firmware push on a blocked powered side but still caches the target', async () => {
```
ng, line 552: **🟢 Merge dropped dev's no-override assertion** (Testing)

origin/dev's equivalent blocked-target test asserted `expect(registerManualOverride).not.toHaveBeenCalled()`; the merged version omits it. Production ordering is still correct (guard check precedes `registerManualOverride`), but the coverage that a rejected write must not open a manual-override anti-thrash window regressed.

**Suggested fix:** restore the one-line assertion.

<details><summary>Review context</summary>

- Source: Codex Optimizer
- Skeptic verdict: Agree (Claude 95, Codex 100 — verified against origin/dev)
- Confidence: Very high
</details>

ng:
**F1 resolution (disputed finding, review 4751770500): option (c) — requested-intent contract, kept.**

`lastTargetF` deliberately records what the user *asked for*, not what firmware confirmed. A guard-rejected temperature write keeps the staged value; iOS Home shows the slider revert (HAP freezes the characteristic at oldValue on rejection), and the staged value applies only on the next explicit power-on after the side is re-enabled.

Why this is safe enough to keep:
- Re-enabling a blocked side is only possible from the web UI's pump-stall alert card — there is no HomeKit acknowledge path. The staged/visible mismatch will be surfaced on that card (plus a transient snackbar on rejection), which sits exactly at the decision point before any heating is possible. Tracked as `sleepypod-core-19`; ships as a separate PR to keep this one focused.
- The fixes for F11/M1 already on this branch narrowed the disputed window: stale ON latches now drop on the first status frame after a trip, after which slider drags on the blocked side stage silently with no visible/hidden divergence.

Implemented on this branch (c20e7ec): contract documented in `sideController.ts`, plus an integration test through the real hap-nodejs characteristic layer pinning the full story — `-70412` rejection at the HAP boundary, visible value frozen, retained intent applied on the next power-on.

Also addressed from this review: F2/F3/F4 (unified guard assert throwing `NOT_ALLOWED_IN_CURRENT_STATE`, REST-aligned wording), F5 (ADR 0022 full ingress inventory), F6–F10 (test hardening incl. real held-lock queued-trip coverage), F11 (stale ON latch dropped when guard-blocked + observed-OFF), M1 (guard rejection forces intent OFF). F12/F13 remain tracked as `sleepypod-core-10` / `sleepypod-core-17`.

PR #671, opened 2026-07-21
fix(streaming): include pump-stall notices in WebSocket status frames

ng:
## Problem

Adversarial review of PR #669, finding S3: the pump-stall banner is driven by `pumpStallNotifications` in device status, but only the tRPC HTTP path (`device.getStatus` in `src/server/routers/device.ts`) includes it via `getAllPumpStallNotices()`. The WebSocket `deviceStatus` producers omit the field, and `useDeviceStatus` prefers live WS frames over HTTP wholesale — so while the WS stream is healthy the stall banner never renders even though the guard tripped.

## Fix

Include `pumpStallNotifications` in every WS `deviceStatus` frame, mirroring the HTTP path's conditional-inclusion semantics (key present only when at least one side has an active notice, so the wire shape stays zod-compatible):

- `src/hardware/dacMonitor.instance.ts` — the `status:updated` broadcast (2s poll path)
- `src/streaming/broadcastMutationStatus.ts` — the mutation-overlay broadcast

No hook change needed: `useDeviceStatus` already maps `frame.pumpStallNotifications` and `DeviceStatusFrame` already declares the optional field.

## Test plan

- Pinning tests for both producers (Stryker): field present when a side has a notice, absent when both sides are null — `src/hardware/tests/dacMonitor.instance.test.ts`, `src/streaming/tests/broadcastMutationStatus.test.ts`
- `pnpm tsc && pnpm lint && pnpm test` clean (3197 passed; pre-push hook re-ran gates)
- Manual: trip the pump-stall guard on a pod with the WS stream connected; the stall banner should render from the next frame without waiting for an HTTP poll

## Review follow-ups (adversarial review)

- **Heads-up:** the MQTT/Home Assistant bridge mirrors WS `deviceStatus` frames verbatim to the retained `device-status` topic, so `pumpStallNotifications` now appears there during a stall — new field for HA consumers, intended.
- Applied from the review (5 commits): stall-guard hardware writes (trip cutoff, retry, auto-recovery) serialized through `withSideLock` with the notice/alert/DB mirror published before the cutoff await; per-side stall-guard frame processing coalesced to one in-flight pass; left-side notice inclusion pinned in all three status producers; `DeviceStatusFrame` sides now typed as `SideStatus & { isAlarmVibrating }`; stale status-stream JSDoc corrected.

<details open>

```bash
sudo sp-update fix/pump-stall-ws-notices
```

```bash
curl -fsSL https://raw.githubusercontent.com/sleepypod/core/fix/pump-stall-ws-notices/scripts/install \
  | sudo bash -s -- --branch fix/pump-stall-ws-notices
```

🎯 **Pin to this exact build** ([run 29959976457](https://github.com/sleepypod/core/actions/runs/29959976457) · `a69ca26`) — fresh-pod path, locked to one CI artifact for reproducible review:
```bash
curl -fsSL https://raw.githubusercontent.com/sleepypod/core/a69ca26985b1f7bd43e8b0f32d22fa7b881936a4/scripts/install \
  | sudo bash -s -- \
      --branch fix/pump-stall-ws-notices \
      --artifact-url 'https://nightly.link/sleepypod/core/actions/runs/29959976457/sleepypod-core.zip'
```

🧊 Artifact: [`sleepypod-core`](https://github.com/sleepypod/core/actions/runs/29959976457/artifacts/8545601539) · self-destructs in 30 days 💥
</details>

## Summary by CodeRabbit

* **New Features**
  * Device status broadcasts now include pump-stall notifications only when active, with the affected side populated.
  * The status API now exposes pump-stall notification details when present.

* **Bug Fixes**
  * Improved pump-stall guard handling to avoid overlapping operations per side and ensure cutoff/restore actions run in order.

* **Tests**
  * Added coverage for left-only, right-only, and no-notice broadcast payloads, plus concurrency/queuing scenarios and router status behavior.

* **Documentation**
  * Refreshed inline documentation to reflect adaptive device-status polling behavior.

### DIFF
--- a/src/hardware/tests/dacMonitor.instance.test.ts
+++ b/src/hardware/tests/dacMonitor.instance.test.ts
@@ -90,6 +90,7 @@ const cancelSnoozeMock = vi.fn<(side: 'left' | 'right') => void>()
 const resetPrimingStateMock = vi.fn()
 const trackPrimingStateMock = vi.fn<(priming: boolean) => void>()
 const getPrimeCompletedAtMock = vi.fn(() => null as number | null)
+const getAllPumpStallNoticesMock = vi.fn<() => { left: unknown, right: unknown }>(() => ({ left: null, right: null }))
 const getAlarmStateMock = vi.fn(() => ({ left: false, right: false }))
 const getSnoozeStatusMock = vi.fn<(side: 'left' | 'right') => { active: boolean, snoozeUntil: number | null }>(
   () => ({ active: false, snoozeUntil: null }),
@@ -128,6 +129,10 @@ vi.mock('../primeNotification', () => ({
   trackPrimingState: (priming: boolean) => trackPrimingStateMock(priming),
   resetPrimingState: () => resetPrimingStateMock(),
   getPrimeCompletedAt: () => getPrimeCompletedAtMock(),
+}))
+
+vi.mock('../pumpStallNotification', () => ({
+  getAllPumpStallNotices: () => getAllPumpStallNoticesMock(),
 }))
 
 vi.mock('../snoozeManager', () => ({
@@ -197,6 +202,7 @@ describe('hardware/dacMonitor.instance', () => {
     resetPrimingStateMock.mockClear()
     trackPrimingStateMock.mockClear()
     getPrimeCompletedAtMock.mockReset().mockReturnValue(null)
+    getAllPumpStallNoticesMock.mockReset().mockReturnValue({ left: null, right: null })
     getAlarmStateMock.mockReset().mockReturnValue({ left: false, right: false })
     getSnoozeStatusMock.mockReset().mockReturnValue({ active: false, snoozeUntil: null })
     broadcastFrameMock.mockClear()
@@ -592,6 +598,38 @@ describe('hardware/dacMonitor.instance', () => {
       const lastFrame = broadcastFrameMock.mock.calls.at(-1)?.[0] as Record<string, unknown> | undefined
       expect(lastFrame).toBeDefined()
       expect(lastFrame?.primeCompletedNotification).toEqual({ timestamp: 123_456 })
+    })
+
+    it('status:updated includes pumpStallNotifications when one side has an active notice', async () => {
+      const mod = await freshModule()
+      await mod.getDacMonitor()
+      await flushMicrotasks()
+      const monitor = monitorInstances[0]
+      const notice = { alertId: 42, trippedAt: 1_700_000_000, rpm: 0, restore: null }
+      getAllPumpStallNoticesMock.mockReturnValue({ left: null, right: notice })
+
+      const status: DeviceStatus = parseDeviceStatusMock('raw')
+      monitor.emit('status:updated', status)
+      await flushMicrotasks()
+
+      const lastFrame = broadcastFrameMock.mock.calls.at(-1)?.[0] as Record<string, unknown> | undefined
+      expect(lastFrame).toBeDefined()
+      expect(lastFrame?.pumpStallNotifications).toEqual({ left: null, right: notice })
+    })
+
+    it('status:updated omits pumpStallNotifications when both sides are null', async () => {
+      const mod = await freshModule()
+      await mod.getDacMonitor()
+      await flushMicrotasks()
+      const monitor = monitorInstances[0]
+
+      const status: DeviceStatus = parseDeviceStatusMock('raw')
+      monitor.emit('status:updated', status)
+      await flushMicrotasks()
+
+      const lastFrame = broadcastFrameMock.mock.calls.at(-1)?.[0] as Record<string, unknown> | undefined
+      expect(lastFrame).toBeDefined()
+      expect(lastFrame && 'pumpStallNotifications' in lastFrame).toBe(false)
     })
 
     it('status:updated does NOT crash when trackPrimingState throws', async () => {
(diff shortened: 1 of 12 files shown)


src/hardware/tests/dacMonitor.instance.test.ts
```diff
@@ -594,6 +600,38 @@ describe('hardware/dacMonitor.instance', () => {
       expect(lastFrame?.primeCompletedNotification).toEqual({ timestamp: 123_456 })
     })
 
+    it('status:updated includes pumpStallNotifications when one side has an active notice', async () => {
```
ng, line 603: **🟢 New pump-stall tests never exercise the left side as the active side** (Testing)

Both new positive tests (here and in `src/streaming/tests/broadcastMutationStatus.test.ts:227`) activate only the right-side notice. A regression — or a Stryker mutant — gating inclusion on only `stallNotices.right` would pass both suites, letting left-side stalls vanish from WS frames undetected.

**Suggested fix:** Add a left-only case per file (mirror the right case with sides swapped). Note: the right-only shape is copied from the pre-existing `device.test.ts:246-253` pattern, so the tRPC path has the same gap — a follow-up covering all 3 sites would close the class.

<details>
<summary>Review context</summary>

- Source: Optimizer (cross-vendor: sonnet + codex gpt-5.5 independently)
- Skeptic verdict: Agree with modifications (confidence 65)
- Haiku isolation score: 100
- Confidence: High (cross-vendor consensus)
</details>

ng (commented):
**Adversarial review — standard depth + Codex cross-vendor sidecar** (escalation score 1; change types: general, test)

**Verdict: patch is correct — Approve.** Both Optimizers (Claude Sonnet, Codex gpt-5.5) and the Skeptic independently reached the same verdict. Mechanical gates on the updated branch (post dev merge `c2448e2`): tsc clean, lint clean, 3279 tests passed / 1 skipped. Grep confirms the two patched sites are the only server-side producers of `type: 'deviceStatus'` WS frames, so the fix is complete for the stated bug.

**In-PR findings (2, both 🟢 Minor):** left-side test asymmetry (inline below — cross-vendor consensus, worth doing) and a disputed S3-completeness claim (inline below — Skeptic refuted it as scope creep).

**Behavior note worth a line in the PR body:** the MQTT/Home Assistant bridge mirrors WS `deviceStatus` frames verbatim to its retained `device-status` topic (`src/streaming/mqttBridge.ts:791`), so `pumpStallNotifications` will now start appearing there on a stall — benign and arguably desirable, but a new field HA consumers haven't seen before.

**Pre-existing issues surfaced (🟣, not this PR's fault, verified by Skeptic + isolation scorer):**
1. `pumpStallGuard.ts:362` — trip/retry/recovery hardware writes bypass `withSideLock` (already tracked as ygg `sleepypod-core-10`)
2. `deviceStateSync.ts:351` — every frzHealth frame fires an untracked `runStallGuard`; concurrent frames can enqueue duplicate guard transitions on the sequential DAC transport
3. `pumpStallGuard.ts:333-362` — `state.blocked` is set before the cutoff await but the user-visible notice is created only after it returns; a 30s DAC timeout window with rejections and no banner
4. `useSensorStream.ts:123` — `DeviceStatusFrame` types side temps non-null and omits `heatingDuration`, contradicting the `SideStatus` payload actually spread into frames
5. Stale JSDoc: fixed-2s polling claims vs adaptive 1/2/5s intervals; `broadcastMutationStatus` "all writers" claim vs actual callers

Review artifacts: `.reviews/fix-pump-stall-ws-notices/` (gitignored). Codex Skeptic sidecar was stopped mid-run — Skeptic pass is Claude-only; Optimizer pass had full cross-vendor coverage.

src/streaming/broadcastMutationStatus.ts
```diff
@@ -42,6 +44,7 @@ export function broadcastMutationStatus(
       waterLevel: lastStatus.waterLevel,
       isPriming: lastStatus.isPriming,
       ...(primeCompletedAt && { primeCompletedNotification: { timestamp: primeCompletedAt } }),
+      ...((stallNotices.left || stallNotices.right) && { pumpStallNotifications: stallNotices }),
```
ng, line 49: **⚠️ Disputed: stall-notice *transitions* still ride the next poll frame** (Completeness)

Codex Optimizer claimed PR #669's finding S3 also required an immediate broadcast on trip/ack/dismiss — notice changes between polls reach clients only on the next DacMonitor frame (≤5s), and a fresh WS frame beats any HTTP refetch. The Claude Skeptic **disputes** this (confidence 80): the actual PR #669 review text only documents that WS frames omit the field — which is exactly what this PR fixes; no transition-broadcast requirement exists there.

**Recommendation:** treat as an optional follow-up (call `broadcastMutationStatus()` after final notice transitions), not as a gap in this PR.

<details>
<summary>Review context</summary>

- Source: Optimizer (codex gpt-5.5 only)
- Skeptic verdict: Disagree (confidence 80) — reclassify as new suggestion
- Model consensus: split (cross-vendor disagreement)
- Confidence: Low as a defect; plausible as an enhancement
</details>

ng:
Review follow-up applied (c2448e2..c86432d): the left-side test gap (inline finding 1) is fixed in d070445 across all three producers, and the surfaced pre-existing issues are fixed — guard writes serialized through `withSideLock` + notice published before the cutoff await (f567b5e), per-side guard coalescing (4db3ea5), `DeviceStatusFrame`/`SideStatus` type alignment (35542d2), JSDoc corrections (c86432d). The disputed S3-transition-broadcast item was intentionally not implemented. Gates: tsc/lint clean, 3289 tests passing. ygg task sleepypod-core-10 closed.

ng:
✅ Fixed - both nitpicks addressed in 98bd9fe / a69ca26: `queueStallGuard` now releases the in-flight flag in a `finally`, and the omission assertion guards against an absent frame.

PR #672, opened 2026-07-21
fix(hardware): enforce pump-stall guard at every energizing ingress

ng:
## Problem

The adversarial review of PR #669 (findings **M13** and **S4**) confirmed that the pump-stall guard's interlock was only enforced at three ingresses (`device.setTemperature`, `device.setPower`, `temperatureKeepalive`). ADR 0022's parking contract says a tripped side stays de-energized until a human acknowledges — but every other energizing path wrote straight to hardware:

- **M13 — bypassing ingresses:** scheduler jobs (temperature, power-on, alarm temperature, away-return, run-once set points), the Automations engine executor, HomeKit setters (thermostat/power switch via `sideController`), gesture actions (temperature taps, power toggle), and `runOnce.start`'s immediate power-on. Any of these could silently re-energize a side parked against a stalled pump — e.g. a 3am scheduled power-on re-heating stagnant water, which is exactly the incident ADR 0022 exists to prevent.
- **S4 — TOCTOU in device router:** the guard check ran before the debounce window (`setTemperature`) / before side-lock acquisition (`setPower`). A trip landing between the check and the hardware write let a stale queued command re-energize the side.

## Fix

Every energizing ingress now consults `shouldBlock` from `src/hardware/pumpStallGuard` (de-energizing writes are never blocked — powering off is the guard's own safe direction):

- `src/scheduler/jobManager.ts` — `runTemperatureJob`, `runPowerOnJob` (inside the side lock), `runAlarmJob` (temperature skipped, **vibration still fires**), away-return power-on (awayMode still cleared in DB), run-once set-point jobs. All degrade to a `console.warn` + skip; the scheduler loop never throws.
- `src/automation/engine.ts` — new injected `pumpStallShouldBlock` dep (wired in `instance.ts`); the check runs inside the side lock immediately before the write and records a `skipped: 'pump-stall'` action in the automation_runs audit log. `setPower(off)` still passes.
- `src/homekit/accessories/sideController.ts` — `setTargetTemperature` skips the push (target stays staged, same semantics as adjusting while OFF); `setSidePowerOn` throws inside the lock so the existing catch rolls `intendedPower` back and iOS Home reverts the toggle instead of showing a phantom ON; `setSidePowerOff` is untouched.
- `src/hardware/gestureActionHandler.ts` — temperature taps and power-on toggles skip with a warn; power-off toggles still work.
- `src/server/routers/runOnce.ts` — `start` fails fast with `PRECONDITION_FAILED` before cancelling the existing session or powering on.
- `src/server/routers/device.ts` (S4) — `setTemperature`/`setPower` now **re-check inside `withSideLock`** immediately before the hardware write, in addition to the existing early fast-fail check.

Deliberately not guarded: all `setPower(side, false)` paths (scheduler power-off/cleanup, away-start, auto-off watcher, HomeKit off, gesture power-off, automation `setPower(off)`), alarm vibration (`setAlarm` — not thermal), and daily priming (circulation-only maintenance; no heat against a stalled pump, and a prime can help clear one). `pumpAlerts.acknowledge` restores through the guarded `device` router after clearing the guard, per the ADR's acknowledgement path.

## Test plan

- Unit tests added per ingress (Stryker-pinning: exact warn strings, blocked-vs-allowed direction, and the S4 in-lock re-check):
  - `src/scheduler/tests/jobManager.test.ts` — temp/power-on/alarm/away-return/run-once blocked variants
  - `src/automation/tests/engine.test.ts` — setTemperature and setPower(on) skipped, setPower(off) allowed, per-side fan-out
  - `src/homekit/tests/sideController.test.ts` — staged-but-not-pushed target, power-on rejection + intent rollback, power-off allowed
  - `src/hardware/tests/gestureActionHandler.test.ts` — temp tap and power-on skipped, power-off allowed
  - `src/server/routers/tests/device.test.ts` — guard trip during the debounce window / lock queue blocks the queued command
  - `src/server/routers/tests/runOnce.test.ts` — start fails fast without cancelling the prior session
- `pnpm tsc && pnpm lint && pnpm test` green (3210 passed, 0 failed); pre-push hook re-ran the gates on push.
- Manual verification on a pod: trip the guard (or force `blocked` via the debug console), then confirm (1) a due temperature schedule logs the skip and sends nothing (`journalctl -u frank` shows no DAC command), (2) the Home app power toggle reverts with an error, (3) alarm vibration still fires at wake time, (4) acknowledging the alert restores the pre-stall target through the normal path.

<details open>

```bash
sudo sp-update fix/pump-stall-guard-enforcement
```

```bash
curl -fsSL https://raw.githubusercontent.com/sleepypod/core/fix/pump-stall-guard-enforcement/scripts/install \
  | sudo bash -s -- --branch fix/pump-stall-guard-enforcement
```

🎯 **Pin to this exact build** ([run 29890757337](https://github.com/sleepypod/core/actions/runs/29890757337) · `6afd2d6`) — fresh-pod path, locked to one CI artifact for reproducible review:
```bash
curl -fsSL https://raw.githubusercontent.com/sleepypod/core/6afd2d669eb443f3c11823b6a4e1407612ccb233/scripts/install \
  | sudo bash -s -- \
      --branch fix/pump-stall-guard-enforcement \
      --artifact-url 'https://nightly.link/sleepypod/core/actions/runs/29890757337/sleepypod-core.zip'
```

🧊 Artifact: [`sleepypod-core`](https://github.com/sleepypod/core/actions/runs/29890757337/artifacts/8518168036) · self-destructs in 30 days 💥
</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Pump-stall protection now blocks temperature changes and power-on activation across automation, scheduled jobs, gesture controls, HomeKit, and server routes, with warnings and “skipped” outcomes when blocked.
  * Power-off operations continue to work while a side is blocked.
  * Queued hardware commands are rechecked inside per-side locks to prevent stale energizing after the guard activates.
  * Power-on rollbacks are corrected (e.g., HomeKit toggles revert properly), including run-once mid-flight rechecks.
  * Manual override anti-thrash no longer persists after expiry, and keepalive ticks are now serialized per side.

### DIFF
--- a/docs/adr/0016-raw-command-execution.md
+++ b/docs/adr/0016-raw-command-execution.md
@@ -25,6 +25,16 @@ Add a `POST /device/execute` endpoint that accepts a hardware command name and o
 - Can cause unexpected hardware state if misused
 - Not covered by the standard safety/debounce mechanisms
 
+## Amendment (2026-07-21)
+
+Energizing opcodes (`SET_TEMP`, `LEFT_TEMP_DURATION`, `RIGHT_TEMP_DURATION`,
+`TEMP_LEVEL_LEFT`, `TEMP_LEVEL_RIGHT` — by name or numeric opcode) are now
+gated on the pump-stall guard (ADR 0022) and serialized through the side
+lock. A level/duration write of `0` (the power-off direction) is never
+blocked, and unknown opcodes still pass through unguarded for probing. To
+deliberately re-energize a tripped side, acknowledge the alert first
+(`pumpAlerts.acknowledge`).
+
 ## Disclaimer
 
 This is a power user feature. It is unsupported, undocumented beyond this ADR, and carries no guarantees. Misuse can lead to unexpected hardware behavior. The sleepypod project assumes no liability for issues arising from raw command execution.
--- a/src/automation/engine.ts
+++ b/src/automation/engine.ts
@@ -61,6 +61,9 @@ export interface AutomationEngineDeps {
   clock: () => { nowMinutes: number, dayOfWeek: DayOfWeek, dateKey?: string }
   getHardware: () => HardwareWriter
   withSideLock: <T>(side: Side, fn: () => Promise<T>) => Promise<T>
+  /** Pump stall guard gate — energizing writes are skipped while it blocks
+   * the side (ADR 0022: a tripped side stays parked until acknowledged). */
+  pumpStallShouldBlock: (side: Side) => boolean
   broadcast: (side: Side, overlay: Record<string, unknown>) => void
   markMutated: (side: Side) => void
   loadRules: () => Promise<AutomationRule[]>
@@ -406,8 +409,16 @@ export class AutomationEngine {
       return { kind: action.kind, side, dryRun: true, raw, temp, clamped, on: action.kind === 'setPower' ? action.on : undefined }
     }
 
-    // Real write through the shared, serialized hardware path.
+    // Real write through the shared, serialized hardware path. The stall-guard
+    // check runs inside the lock so a trip while this write is queued still
+    // blocks it; power-off is the safe direction and is never blocked.
+    const energizing = action.kind === 'setTemperature' || action.on
+    let stallBlocked = false
     await this.deps.withSideLock(side, async () => {
+      if (energizing && this.deps.pumpStallShouldBlock(side)) {
+        stallBlocked = true
+        return
+      }
       const hw = this.deps.getHardware()
       await hw.connect()
       if (action.kind === 'setTemperature') {
@@ -427,6 +438,10 @@ export class AutomationEngine {
         else if (!action.on) this.lastAsserted[side] = undefined
       }
     })
+    if (stallBlocked) {
+      console.warn(`[automation] skipped ${action.kind}: pump stall guard blocks ${side}`)
+      return { kind: action.kind, side, skipped: 'pump-stall', raw, temp, clamped }
+    }
     rt.actionTimes.push(now)
     return { kind: action.kind, side, sent: true, raw, temp, clamped, on: action.kind === 'setPower' ? action.on : undefined }
   }
--- a/src/automation/engine.ts
+++ b/src/automation/engine.ts
@@ -62,7 +62,8 @@ export interface AutomationEngineDeps {
   getHardware: () => HardwareWriter
   withSideLock: <T>(side: Side, fn: () => Promise<T>) => Promise<T>
   /** Pump stall guard gate — energizing writes are skipped while it blocks
-   * the side (ADR 0022: a tripped side stays parked until acknowledged). */
+   * the side (ADR 0022: a tripped side stays parked until acknowledgment
+   * or successful opt-in auto-recovery). */
   pumpStallShouldBlock: (side: Side) => boolean
   broadcast: (side: Side, overlay: Record<string, unknown>) => void
   markMutated: (side: Side) => void
@@ -163,6 +164,10 @@ export class AutomationEngine {
    */
   registerManualOverride(side: Side): void {
     this.manualOverrideUntil[side] = this.deps.now() + AUTOMATION_MANUAL_OVERRIDE_MS
+    // The user changed the setpoint out from under us — the anti-thrash
+    // baseline no longer reflects hardware, and keeping it would suppress
+    // the first re-assertion after the override window expires.
+    this.lastAsserted[side] = undefined
   }
 
   /** Flip the global kill-switch. `false` suspends all evaluation immediately. */
--- a/src/automation/instance.ts
+++ b/src/automation/instance.ts
@@ -12,6 +12,7 @@ import { db } from '@/src/db'
 import { automationRuns, automations, deviceSettings, runOnceSessions } from '@/src/db/schema'
 import { getSharedHardwareClient } from '@/src/hardware/dacMonitor.instance'
 import { markSideMutated } from '@/src/hardware/deviceStateSync'
+import { shouldBlock as pumpStallShouldBlock } from '@/src/hardware/pumpStallGuard'
 import { withSideLock } from '@/src/hardware/sideLock'
 import { broadcastMutationStatus } from '@/src/streaming/broadcastMutationStatus'
 import { AutomationEngine } from './engine'
@@ -117,6 +118,7 @@ export async function getAutomationEngine(): Promise<AutomationEngine> {
         clock: () => clockInTimezone(activeTimezone, new Date()),
         getHardware: () => getSharedHardwareClient(),
         withSideLock,
+        pumpStallShouldBlock,
         broadcast: (side, overlay) => broadcastMutationStatus(side, overlay),
         markMutated: markSideMutated,
         loadRules,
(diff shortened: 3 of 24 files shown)


ng:
## Adversarial review — `/adversarial-review:run` (standard depth + Codex sidecar)

**Reviewers:** Sonnet Optimizer + Codex (gpt) Optimizer sidecar → merged → Sonnet Skeptic (Codex Skeptic was stopped mid-run; Optimizer-side cross-vendor coverage only). Haiku isolation scoring ran on all 8 confirmed findings (all ≥ 75, no downgrades). Escalation score 3 (baseline +1, api +2) → standard depth. Mechanical checks: tsc ✅, lint ✅, 3244 tests ✅.

### Fixed in this PR (pushed as `8b9fbb8`)

| # | Finding | Provenance | Verdict |
|---|---------|------------|---------|
| F1 | `runOnce.start` TOCTOU — guard checked only at mutation entry; session-cancel + settings awaits sit before the unlocked `setPower(true)` | sonnet + codex (cross-vendor) | Skeptic ✅ 85, Haiku 80 |
| F3 | Away-return power-on ran outside `withSideLock` with no in-lock re-check | codex | Skeptic ✅ 85, Haiku 80 |

F1: write-time re-check inside the `withHardwareClient` callback, same `PRECONDITION_FAILED` as device.ts. F3: power-restore now wrapped in `withSideLock` with the guard checked inside, matching `runPowerOnJob`. Both have pinning tests (incl. a real lock-contention test that trips the guard while the job is queued).

### Deferred — needs its own PR

**F2 (Major, Skeptic 🔄 90, Haiku 82): the guard's own cutoff isn't serialized through `withSideLock`** (`pumpStallGuard.ts` `trip()`/`onFrame` retry), and several ingress gates run before `await client.connect()`. The Skeptic mechanically reproduced the race (5/5 runs: OFF lands, then the stale ON re-energizes) — but **only when `connect()` does real work** (reconnect); in steady-state no-op connect the sequence is safe. The unlocked cutoff is pre-existing and untouched by this diff. Recommended fix (one change closes every ingress at once): serialize `trip()`'s cutoff + retries through `withSideLock`. Tracked as a follow-up — safety-path change needing its own deadlock analysis and tests.

### Worth a second look (Minor, confirmed, not auto-fixed)

- **F4** `device.ts:344` — optimistic `isPowered=true` UPDATE lands before the new in-lock re-check; late rejection leaves DB/UI claiming energized (Skeptic ✅ 70, Haiku 82).
- **F5** `device.ts:341/:453` — `registerManualOverride` runs before the final guard check; a rejected command still suspends autopilot. Same PR orders it *after* the check in sideController/gestureActionHandler (Skeptic ✅ 80, Haiku 75).
- **F6** `helpers.ts:44` — retry catch wraps a `PRECONDITION_FAILED` thrown on the retry attempt into a 500; one-line `instanceof TRPCError` rethrow (Skeptic ✅ 90, Haiku 95).
- **F7** — the four new ingress test files only test blocked-from-start; none pins the mid-flight trip (device.test.ts's `mockReturnValueOnce(false).mockReturnValue(true)` pattern). Partially addressed for runOnce/away-return by `8b9fbb8` (Haiku 78).
- **F8** `device.test.ts:517/:707` — "re-checks inside the lock" tests use call-count mocks, no real contention; they'd pass if the re-check moved outside the lock (Haiku 78).

### Nits
- F9: the `PRECONDITION_FAILED` throw block is now duplicated 6×— optional `assertPumpStallNotBlocked(side)` helper.
- F10: `engine.ts:65` comment says parked "until acknowledged"; ADR 0022 also allows opt-in auto-recovery.

### Pre-existing (🟣, not this PR's fault)
P1 keepalive reissue unlocked (`temperatureKeepalive.ts:50`) · P2 `trip()` mirrors `isPowered=false` so cutoff-retry/auto-recovery frames bail early at `pumpStallGuard.ts:140` — retry logic normally unreachable · P3 away-start power-off unlocked, no `markSideOff` · P4 runOnce cancels sessions before the 14h validation · P5 runOnce powers on before session insert · P6 `registerManualOverride` leaves `lastAsserted` stale · P7 `device.execute` bypasses the guard (documented debug escape hatch).

### Verdict
**Comment — approve direction.** Both Skeptic and synthesis agree the patch is a strict safety improvement with no regressions (full suite green before and after). The "every energizing ingress" claim required F1/F3 (now fixed); F2 remains the one open Major and is tracked as a follow-up.

ng:
## Review feedback round 2 — Minor/Nit/pre-existing findings addressed

All remaining findings from the adversarial review are now fixed and pinned (suite 3246 → 3265 green; tsc/lint clean):

| Finding | Fix | Commit |
|---|---|---|
| F4 late guard rejection left DB `isPowered=true` | parked mirror restored + broadcast on PRECONDITION_FAILED | 5efbeeb |
| F5 rejected commands registered a manual override | `registerManualOverride` moved after the in-lock recheck (matches sideController/gesture pattern) | 5efbeeb |
| F6 retry masked PRECONDITION_FAILED as 500 | retry catch rethrows `TRPCError` | 5efbeeb |
| F9 (nit) PRECONDITION_FAILED duplicated 5× | shared `assertPumpStallNotBlocked` helper | 5efbeeb |
| F10 (nit) comment omitted auto-recovery | wording fixed | 9523e95 |
| F7 tests missed mid-flight trips | queued-trip tests added for engine, gestures, scheduler power-on/temp jobs | 0d9b144 / 9523e95 / 72a1a29 |
| F8 lock tests created no real contention | real held-lock + flip-while-queued tests for `setTemperature`/`setPower` | 5efbeeb |
| P1 keepalive reissue unlocked | whole tick under `withSideLock`, checks at the write | 44e241c |
| P2 cutoff retry / auto-recovery unreachable post-trip | blocked sides keep processing `expectedActive=false` frames | b8049ea |
| P3 away-start power-off unlocked, no `markSideOff` | runPowerOffJob protocol: lock + DB-off-first | 0d9b144 |
| P4 run-once cancelled sessions before validation | validation reordered ahead of cancellation | 1c65010 |
| P5 insert failure orphaned an energized side | compensating power-off on insert failure | 1c65010 |
| P6 stale `lastAsserted` after manual override | anti-thrash baseline reset on override | 9523e95 |
| P7 `device.execute` guard bypass | intentional, documented POWER USER escape hatch — left as designed | — |

Still deferred: **F2** (serialize the guard's own `trip()` cutoff through `withSideLock`) — tracked as `sleepypod-core-10` (P1); pre-existing root cause with its own deadlock analysis. `sleepypod-core-9` (auto-recovery's hard-coded 28800s restore) also remains open and is now *reachable* behavior after the P2 fix, so it gains priority.

ng:
## Review feedback round 3 — P7 addressed

**P7** (`device.execute` guard bypass) is now closed rather than left as designed — `6afd2d6`:

- Energizing opcodes (`SET_TEMP`, `LEFT/RIGHT_TEMP_DURATION`, `TEMP_LEVEL_LEFT/RIGHT`) are gated on `assertPumpStallNotBlocked`, whether spelled by name **or numeric opcode** (so `'11'` can't sidestep `TEMP_LEVEL_LEFT`).
- Energizing raw writes queue on `withSideLock` with the in-lock re-check, matching every other ingress in this PR. `SET_TEMP` spans both sides — nested left→right, the codebase's only nested acquisition, with the ordering documented.
- A level/duration write of `0` (power-off direction) is never blocked, per ADR 0022. Unknown numeric opcodes still pass through unguarded — the probing escape hatch survives; deliberate re-energizing of a tripped side goes through `pumpAlerts.acknowledge` first.
- ADR 0016 amended; 10 pinning tests added (blocked named + numeric forms, both-sides SET_TEMP, level-0 exemption, unguarded passthrough, mid-flight trip re-check, real held-lock contention).

Branch also rebased onto latest `dev`. Suite 3265 → 3275 green; tsc/lint clean.

PR #675, opened 2026-07-26
fix(biometrics): cap archive by folder size; check every fs in sp-update preflight

paulkearney:
## What happened

A pod filled its 15 GB eMMC to 100% — 1983 files / 13 GB of cold archive — and `sp-update` then failed with `No space left on device`. Two independent gaps let that happen, and a third made it invisible.

The pruner had been failing **every 15 minutes for a month**:

```
sleepypod-biometrics-pruner[462957]: sort: write failed: 'standard output': Broken pipe
systemd[1]: Main process exited, code=exited, status=2/INVALIDARGUMENT
```

That specific `head`-in-a-pipeline SIGPIPE is already fixed on `dev` (`awk NR==1`) — the pod was simply running a month-old build. This PR addresses what's left.

## Changes

**Pruner: cap the folder, not the disk percent.** The bound was whole-disk percent with `-le`, so it stopped as soon as usage was *at* 80% and never recovered headroom. On a 15 GB eMMC that blesses a ~12 GB archive and leaves the disk permanently pinned at 80%. Since `archive-push` is copy-only, the pruner is the only thing bounding the archive, so it now caps the folder at `MAX_ARCHIVE_MB` (default 6 GiB), with the 80% ceiling kept only as a backstop against growth from *outside* the archive.

**Pruner: fail loudly.** When neither bound can be reached it exits non-zero, so a disk filling from elsewhere shows up in `systemctl --failed` instead of a silent give-up inside a success log. The silent give-up is a large part of why this went unnoticed for a month.

**Pruner: materialise the list once.** A catch-up run can be ~1300 removals, and re-deriving the oldest entry with `ls -1tr` over ~2000 entries per deletion is O(n²) on eMMC. Both existing portability choices are deliberately preserved — `find -printf` silently no-ops on busybox and `df --output=pcent` is rejected outright, so this keeps `ls -1tr`, `du -k`, and `df -P`.

**`sp-update` pre-flight: check every filesystem it writes to.** It measured only `$INSTALL_DIR`, but an update writes to three volumes:

| path | filesystem | holds |
|---|---|---|
| `$INSTALL_DIR` | rootfs (5.9 GB) | source tree + `.next` |
| `node_modules` | eMMC (15 GB) | bind-mounted onto `/persistent` |
| `$WORK_DIR` / `$BACKUP_DIR` | tmpfs (`/tmp`) | release tarball + rollback copy |

A healthy rootfs with a full `/persistent` passed the old 300 MB check and then died mid-`pnpm install`, where a bare `ENOSPC` says nothing about which volume ran out. Requirements are now accumulated per mount point (paths sharing a filesystem sum) and reported per filesystem with the actual shortfall.

## Verification

Tested on the affected Pod 3 (bash 5.1, GNU coreutils 9.0):

- 2000 files / 196 MB pruned to a 50 MB cap — oldest-first order confirmed, exit 0; re-run is a no-op
- ~19 s for 1488 removals, 0.05 s in steady state (at `Nice=10` / `IOSchedulingClass=idle`)
- unreachable bound exits 1 with the shortfall reported
- pre-flight: passes when healthy; fails per-filesystem when short; **sums two requirements on a shared filesystem** (9000 + 300 → `/ needs 9300MB`); tolerates a not-yet-created path
- live pod: pruner now logs `Finished` (`pruned=0 archive_mb=2266 used_pct=29%`) instead of `Failed to start`, and `sp-update` completed successfully

ADR 0018 gains a "Pruner constraints (learned the hard way)" section recording all three constraints, so they don't get re-broken.

https://claude.ai/code/session_01DyKwc4SYgs69bNg7iqA8f3

## Summary by CodeRabbit

- **New Features**
  - RAW biometrics archives now use a configurable size cap (6 GiB by default) alongside an 80% whole-disk safety ceiling.
  - Archive cleanup removes the oldest files first when either limit is exceeded.

- **Bug Fixes**
  - System updates now accurately check available space across shared and separate filesystems.
  - Disk-space failures provide clearer information about the affected storage location.

- **Documentation**
  - Updated storage and operational guidance to reflect archive limits and cleanup behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ graph LR
 
 ### Biometrics data flow
 
-The Pod hardware daemon (frankenfirmware) writes raw sensor data continuously as CBOR-encoded binary records into `/persistent/biometrics/*.RAW`, which is a **500 MB tmpfs** to keep ~1 GB/day of writes off the eMMC. An archiver timer gzips files older than 15 minutes into `/persistent/biometrics-archive/` on eMMC (cold storage with a pruner cap of 80% disk usage). See **[ADR 0018](docs/adr/0018-tmpfs-raw-frames.md)** for the firmware-integration rationale, including how `SEQNO.RAW` and state directories are symlinked through so the firmware sees its persistent state unchanged.
+The Pod hardware daemon (frankenfirmware) writes raw sensor data continuously as CBOR-encoded binary records into `/persistent/biometrics/*.RAW`, which is a **500 MB tmpfs** to keep ~1 GB/day of writes off the eMMC. An archiver timer gzips files older than 15 minutes into `/persistent/biometrics-archive/` on eMMC (cold storage; a pruner caps the folder at `MAX_ARCHIVE_MB`, default 6 GiB, with an 80% whole-disk safety ceiling). See **[ADR 0018](docs/adr/0018-tmpfs-raw-frames.md)** for the firmware-integration rationale, including how `SEQNO.RAW` and state directories are symlinked through so the firmware sees its persistent state unchanged.
 
 Independent Python sidecar processes tail the hot tmpfs files, extract signals, and write results to `biometrics.db` (durable on eMMC). The core app never touches raw data — it reads clean rows via tRPC.
 
--- a/docs/adr/0018-tmpfs-raw-frames.md
+++ b/docs/adr/0018-tmpfs-raw-frames.md
@@ -18,7 +18,7 @@ Mount a **500 MB tmpfs at `/persistent/biometrics`**, redirect frankenfirmware's
 frank → /persistent/biometrics/         (tmpfs, hot live RAW)
            ↓ archiver every 15 min: gzip oldest, atomic rename
         /persistent/biometrics-archive/<seqno>.RAW.gz
-           ↓ pruner every 15 min: drop oldest until df < 80%
+           ↓ pruner every 15 min: drop oldest until folder <= MAX_ARCHIVE_MB (and df < 80%)
 ```
 
 ### Firmware integration without a binary patch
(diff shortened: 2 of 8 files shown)


ng:
Reviewed the full PR and pushed fixes in `88df88f`.

Fixed three issues in `sp-update` preflight:

- **Wrong dependency filesystem:** mounting `/persistent` does not mean relocation selects it. On Pod 3 layouts where rootfs is larger, the original check could reject a healthy update because the small `/persistent` lacked 900 MB, or miss insufficient rootfs space. Preflight now uses the relocation helper's target selection, honors `NODE_MODULES_DIR`, and accounts for stale bind mounts that relocation removes.
- **Unchecked rollback filesystem with `TMPDIR`:** rollback always writes to `/tmp/sleepypod-rollback`, while the download workspace follows `TMPDIR`. Both paths now receive separate budgets that sum when they share a filesystem.
- **Shared capacity counted separately:** different bind mount points can consume the same underlying filesystem. Requirements now aggregate by filesystem device ID, following symlinks, while errors retain a readable mount point.

Added 12 executable regression tests covering storage layouts, overrides, shared capacity, rollback/workspace placement, oldest-first archive pruning, the exact disk ceiling, and an unreachable archive cap. Seven preflight tests fail against the original PR and pass after the fixes.

Validation: 12/12 storage tests; full Vitest suite 3,297 passed, 1 skipped; TypeScript; ESLint (one existing warning in `stryker.config.mjs`); both Drizzle schema checks; Bash syntax; ShellCheck at warning severity; `git diff --check`.

The storage tests simulate filesystem layouts and use real temporary archive files; I did not run these changes on a physical Pod. Space budgets remain estimates rather than guarantees for arbitrarily large installations.

PR #677, opened 2026-07-27
fix(hardware): prevent false pump stalls at session expiry and recover remaining sessions

ng:
Normal firmware session expiry could be misclassified as a pump stall, shutting down the bed and raising HomeKit leak alarms. Recovery could also restart a full eight-hour session or replay an expired snapshot, while old alerts could re-block healthy pumps after an update.

This change makes the guard recognize session completion, restore only the remaining session, and resolve stale incidents without repeatedly restarting or blocking the bed.

## Final behavior

- Recognize fresh neutral targets before stale pump duty. When firmware retains a non-neutral target after its countdown reaches zero, retain confirmed expiry from the observed positive countdown. Fresh zero polls keep that evidence valid; a new command, target, countdown, or stale status invalidates it.
- Require sustained low RPM using frame-arrival time, a 10-second dwell floor, frame-gap resets, and a 120-second bilateral dwell for fresh trips.
- Age both manual and automatic restore snapshots. Expired sessions remain off. Bounded recovery probes let powered-off pumps demonstrate healthy RPM, with cutoff retries and incident checks protecting concurrent acknowledgement and dismissal.
- Harden alert rehydration and supersede older incidents. Add a Status-screen alert card with individual and bulk dismissal, while refusing dismissal during an unconfirmed cutoff.
- Repair scheduled-reboot authorization during installation and updates, and make updater firewall restoration fail closed when rule repair fails.

## Validation

- Reproduced the expiry false trip on a live Pod 5 with the guard enabled: the firmware retained target level -20 after a three-minute countdown reached zero; the guard tripped about 50 seconds later.
- Retested final commit `4c38f54` on that pod with the guard enabled. Both sessions expired with retained targets and zero RPM, with no alert or block for more than four minutes beyond expiry. A new left-side session then restarted successfully and reached 1944 RPM.
- Restored the pod's original settings and powered both sides off after testing.
- 280 focused tests passed; the pre-push gate passed 1,581 related tests plus TypeScript, lint, and both schema checks. Production build and GitHub checks passed.

The live test verifies the session-expiry failure on Pod 5. Positive-duty telemetry is covered by automated tests because this pod's frames omit duty. Mechanical-stall recovery and the reporter's Pod 4 were not hardware-tested in this validation.

<details open>

```bash
sudo sp-update fix/pump-stall-restore-remaining-session
```

```bash
curl -fsSL https://raw.githubusercontent.com/sleepypod/core/fix/pump-stall-restore-remaining-session/scripts/install \
  | sudo bash -s -- --branch fix/pump-stall-restore-remaining-session
```

🎯 **Pin to this exact build** ([run 33940841952](https://github.com/sleepypod/core/actions/runs/33940841952) · `4c38f54`) — fresh-pod path, locked to one CI artifact for reproducible review:
```bash
curl -fsSL https://raw.githubusercontent.com/sleepypod/core/4c38f54c1a96de99603246718638ed730124e097/scripts/install \
  | sudo bash -s -- \
      --branch fix/pump-stall-restore-remaining-session \
      --artifact-url 'https://nightly.link/sleepypod/core/actions/runs/33940841952/sleepypod-core.zip'
```

🧊 Artifact: [`sleepypod-core`](https://github.com/sleepypod/core/actions/runs/33940841952/artifacts/9961775304) · self-destructs in 30 days 💥
</details>


ng:
Adversarial follow-up review complete on 9f5d3df.

- Confirmed the pump-stall behavior-bearing blobs are unchanged from field-tested e0b40b7; targeted coverage still passes the 10s dwell floor, 5/15/30 recovery probes, and expired-session no-restore path.
- Fixed OTA sudoers reconciliation under the restricted service PATH by resolving the pod validator at /usr/sbin/visudo (with standard absolute fallbacks).
- Kept reconciliation best-effort under set -e, added deployment regression contracts, and removed both narrow sudoers fragments during uninstall.
- Two Optimizer and two Skeptic passes found no remaining Critical or Major issue.

Verification: full isolated suite 134 files / 3382 passed / 1 skipped; pre-push gate 49 files / 1659 passed; TypeScript, lint, schema validation, build, and shell syntax all pass. Real-device scheduled reboot remains the final field confirmation after OTA.

ng:
Pushed `43f6ff0` to this branch — folds #685's pump-stall false-trip hardening in so both failure modes live in the pump-stall PR:

- **session-end** (real stop, `targetLevel → 0`, duty lags): the `targetLevel`-first reorder already here.
- **glitch** (both frames read 0/0 while pumps spin): #685's 120s bilateral dwell (`d835982`) + frame-arrival clock / gap-reset (`7b983ff`), reimplemented on this branch's guard.

`+54/-1` pumpStallGuard.ts, `+15/-10` deviceStateSync.ts, `+84` test lines. Full pre-push gates green (1724 tests). Bilateral dwell is scoped to the fresh-trip low-run only — the rehydrated energized-stall watch keeps the 10s dwell so a restored block powers off promptly.

Coordination: origin advanced `1d47aac → 43f6ff0`. Anyone holding `codex/pr677-feedback` at `1d47aac` locally should pull/rebase before the next push so this isn't clobbered.

PR #683, opened 2026-08-10
fix(install): write the SSH key where sshd actually reads it

jasonlarue:
## The bug

`scripts/install`'s optional SSH-setup block appended the user's public key to a hardcoded `/root/.ssh/authorized_keys` and, in the same pass, set `PasswordAuthentication no`. Those two halves are individually defensible and jointly a lockout:

- **root's home on this Yocto image is `/home/root`**, not `/root` (poky convention — the same assumption is already baked into `scripts/install:965`, `scripts/bin/sp-maintenance:75`, `scripts/pod/capabilities:90`). sshd resolves the default `AuthorizedKeysFile .ssh/authorized_keys` to `/home/root/.ssh/authorized_keys` and never looks at `/root/.ssh`.
- **free-sleep's `setup_ssh.sh` pins an absolute `AuthorizedKeysFile /home/root/ssh/authorized_keys`.** That directive survives the installer's `sed` edits, so even the corrected relative default would be ignored on a pod that has seen free-sleep.

Either way the key lands somewhere sshd ignores while password auth is switched off. Port 8822 is the pod's only remote entry point, so the result is an unrecoverable lockout — `Permission denied (publickey)` for every user, with recovery requiring a serial console and a U-Boot `init=/bin/bash` boot (or a hardware firmware reset, which wipes the install).

Found the hard way on a real pod: full TCP sweep showed only 111/5355/8822 open, sshd offering `publickey` only, rejecting a known-good key at the offer stage.

## The fix

- **`scripts/lib/ssh-helpers`** (new) — `resolve_authorized_keys_path` asks sshd where it actually reads keys: `sshd -T` first (so an `AuthorizedKeysFile` set in an `Include`d file still wins), then the raw config, then sshd's compiled-in default. Expands `%h`/`%u`, takes the first entry of a list, and resolves relative paths against root's real home from `getent`. Plus `install_authorized_key` (idempotent, StrictModes-safe ownership and modes) and `migrate_stranded_authorized_keys`, which folds keys left behind in `/root/.ssh` by an earlier install into the real file — so re-running the installer *repairs* an existing lockout instead of recreating it.
- **`scripts/install`** — sources the helper (same pattern as `scripts/pod/detect` and `scripts/lib/relocation-helpers`, both already sourced from `$INSTALL_DIR` at that point). Sets `Port` up front, but applies `PermitRootLogin prohibit-password` / `PasswordAuthentication no` / `PermitEmptyPasswords no` **only after a key is verifiably installed**. No valid key means auth settings are left untouched with an explanatory message; a missing helper aborts before touching them.
- **`tests/ssh-helpers.test.ts`** (new, 32 tests) — runs the helper under `bash` with `getent`/`sshd` stubbed to look like a pod, pinning each resolution rule and the migration behaviour.
- **`scripts/README.md`** — documents the real path and the "confirm key auth works before disabling passwords" ordering.

## Notes for review

The tests earned their keep immediately: the first version of the helper aborted the whole installer under `set -euo pipefail` whenever `getent` or `sshd -T` returned non-zero. Hence the explicit `|| true` guards, which are commented as load-bearing.

This is the repo's first shell-level test. It shells out to `bash` from vitest rather than introducing bats; happy to restructure if you'd prefer a different harness.

## Verification

`pnpm tsc` clean; `pnpm lint` clean (one pre-existing warning in `stryker.config.mjs`); `pnpm test` 3321 passed / 1 skipped across 137 files; `bash -n` clean on both scripts.

Not exercised end-to-end on a pod — the pod I found this on is currently locked out by the very bug being fixed.

## Summary by CodeRabbit

* **Improvements**
  * SSH setup detects the effective authorized-keys location, including `/home/root` layouts.
  * Existing keys in legacy locations are migrated automatically without duplication.
  * Key installation validates formatting, enforces secure permissions, and verifies persistence.
  * SSH configuration is validated before restart, with port 8822 applied consistently.
  * Password authentication is disabled only after successful key verification, helping prevent lockouts.
  * Installation and manual setup documentation include path checks, authentication verification, and recovery guidance.

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -1791,17 +1791,26 @@ if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]]; 
   fi
 
   # A directive can be syntactically valid and still not take effect — an
-  # Include or a later duplicate can override it. Confirm what sshd will
-  # actually do before we restart it into that config.
+  # Include, a later duplicate, or a Match block can override it. Confirm
+  # what sshd will actually do before we restart it into that config.
+  #
+  # The auth directives are checked in root's connection context: all three
+  # are Match-able, so a global reading would pass while `Match User root`
+  # re-enabled password auth for exactly the account being hardened. Port
+  # can't appear in a Match block, so it stays global.
   SSH_EXPECTED="port=8822"
   if [ "$SSH_KEY_INSTALLED" = true ]; then
-    SSH_EXPECTED="$SSH_EXPECTED permitrootlogin=prohibit-password"
-    SSH_EXPECTED="$SSH_EXPECTED passwordauthentication=no permitemptypasswords=no"
+    SSH_EXPECTED="$SSH_EXPECTED root:permitrootlogin=prohibit-password"
+    SSH_EXPECTED="$SSH_EXPECTED root:passwordauthentication=no root:permitemptypasswords=no"
   fi
   for expectation in $SSH_EXPECTED; do
+    case "$expectation" in
+      *:*) SSH_CONTEXT="${expectation%%:*}"; expectation="${expectation#*:}" ;;
+      *)   SSH_CONTEXT="" ;;
+    esac
     directive="${expectation%%=*}"
     want="${expectation#*=}"
-    got="$(sshd_effective_value "$directive")"
+    got="$(sshd_effective_value "$directive" /etc/ssh/sshd_config "$SSH_CONTEXT")"
     if [ "$got" != "$want" ]; then
       echo "Error: sshd reports $directive=${got:-<unset>}, expected $want." >&2
       echo "Restoring backup and leaving sshd untouched..." >&2
--- a/scripts/install
+++ b/scripts/install
@@ -1719,16 +1719,20 @@ if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]]; 
     # Without the helpers we can't tell where sshd reads authorized_keys
     # from, and hardening on a guess locks the pod out for good. Leave sshd
     # exactly as it is.
-    echo "Error: $INSTALL_DIR/scripts/lib/ssh-helpers not found — skipping SSH setup." >&2
+    echo "Error: $INSTALL_DIR/scripts/lib/ssh-helpers not found — aborting install." >&2
     echo "sshd was left unchanged. Re-run the installer once the source tree is complete." >&2
     exit 1
   fi
   source "$INSTALL_DIR/scripts/lib/ssh-helpers"
 
-  # Backup sshd_config
+  # Keep the first-ever config as a historical artifact, but restore from a
+  # fresh per-run snapshot — .backup is frozen at the first run, so restoring
+  # it on a re-run reverts to the stock config (port 22, password auth on)
+  # rather than to the state this run started from.
   if [ ! -f /etc/ssh/sshd_config.backup ]; then
     cp /etc/ssh/sshd_config /etc/ssh/sshd_config.backup
   fi
+  cp -f /etc/ssh/sshd_config /etc/ssh/sshd_config.pre-install
 
   # Port is safe to set before we know whether a key will land — it doesn't
   # change who can authenticate. The rest of the hardening waits until the
@@ -1782,7 +1786,7 @@ if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]]; 
   # Validate SSH config before restarting
   if ! sshd -t; then
     echo "Error: sshd_config validation failed. Restoring backup..." >&2
-    cp /etc/ssh/sshd_config.backup /etc/ssh/sshd_config
+    cp /etc/ssh/sshd_config.pre-install /etc/ssh/sshd_config
     exit 1
   fi
 
@@ -1801,7 +1805,7 @@ if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]]; 
     if [ "$got" != "$want" ]; then
       echo "Error: sshd reports $directive=${got:-<unset>}, expected $want." >&2
       echo "Restoring backup and leaving sshd untouched..." >&2
-      cp /etc/ssh/sshd_config.backup /etc/ssh/sshd_config
+      cp /etc/ssh/sshd_config.pre-install /etc/ssh/sshd_config
       exit 1
     fi
   done
@@ -1819,7 +1823,7 @@ if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]]; 
     echo "$SSH_RESULT"
   else
     echo "Error: Failed to restart SSH service" >&2
-    cp /etc/ssh/sshd_config.backup /etc/ssh/sshd_config
+    cp /etc/ssh/sshd_config.pre-install /etc/ssh/sshd_config
     exit 1
   fi
 fi
--- a/scripts/install
+++ b/scripts/install
@@ -547,9 +547,11 @@ preflight_free_sleep_conflict() {
 #   su -                            # switch to root with the passwd you set
 #   bash scripts/install            # run this script
 # The optional SSH-setup block at the end of this script writes your key to
-# /root/.ssh/authorized_keys and re-hardens sshd (port 8822, key-only, no
-# password auth, no root password login). After that you can ssh directly as
-# root with your key.
+# whichever authorized_keys file sshd actually reads for root (resolved by
+# scripts/lib/ssh-helpers — root's home is /home/root here, and free-sleep
+# may have pinned an absolute AuthorizedKeysFile) and only then re-hardens
+# sshd (port 8822, key-only, no password auth, no root password login).
+# After that you can ssh directly as root with your key.
 if [ "$EUID" -ne 0 ]; then
   echo "Error: This script must be run as root (use sudo or su -)" >&2
   echo "" >&2
@@ -1713,32 +1715,57 @@ fi
 if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]]; then
   echo "Configuring SSH..."
 
+  if [ ! -f "$INSTALL_DIR/scripts/lib/ssh-helpers" ]; then
+    # Without the helpers we can't tell where sshd reads authorized_keys
+    # from, and hardening on a guess locks the pod out for good. Leave sshd
+    # exactly as it is.
+    echo "Error: $INSTALL_DIR/scripts/lib/ssh-helpers not found — skipping SSH setup." >&2
+    echo "sshd was left unchanged. Re-run the installer once the source tree is complete." >&2
+    exit 1
+  fi
+  source "$INSTALL_DIR/scripts/lib/ssh-helpers"
+
   # Backup sshd_config
   if [ ! -f /etc/ssh/sshd_config.backup ]; then
     cp /etc/ssh/sshd_config /etc/ssh/sshd_config.backup
   fi
 
-  # Update SSH configuration
+  # Port is safe to set before we know whether a key will land — it doesn't
+  # change who can authenticate. The rest of the hardening waits until the
+  # key is verifiably in place.
   sed -i "s/#\?Port .*/Port 8822/" /etc/ssh/sshd_config
-  sed -i "s/#\?PermitRootLogin .*/PermitRootLogin prohibit-password/" /etc/ssh/sshd_config
-  sed -i "s/#\?PasswordAuthentication .*/PasswordAuthentication no/" /etc/ssh/sshd_config
-  sed -i "s/#\?PermitEmptyPasswords .*/PermitEmptyPasswords no/" /etc/ssh/sshd_config
+
+  # Repair installs from before the authorized_keys path fix, where the key
+  # went to /root/.ssh and sshd (root home = /home/root) never read it.
+  migrate_stranded_authorized_keys root
 
   echo ""
   echo "Enter your SSH public key:"
   read -r SSH_KEY
 
+  SSH_KEY_INSTALLED=false
   if [ -n "$SSH_KEY" ]; then
     # Validate key format
     if echo "$SSH_KEY" | grep -qE '^(ssh-rsa|ssh-ed25519|ecdsa-sha2-nistp[0-9]+) [A-Za-z0-9+/=]+'; then
-      mkdir -p /root/.ssh
-      echo "$SSH_KEY" >> /root/.ssh/authorized_keys
-      chmod 700 /root/.ssh
-      chmod 600 /root/.ssh/authorized_keys
-      echo "SSH key added"
+      AUTHORIZED_KEYS="$(install_authorized_key root "$SSH_KEY")"
+      echo "SSH key added to $AUTHORIZED_KEYS"
+      SSH_KEY_INSTALLED=true
     else
       echo "Warning: Invalid SSH key format, skipping"
     fi
+  fi
+
+  # Only disable password auth once root has a key sshd can actually read.
+  # Hardening without one is an unrecoverable lockout — port 8822 is the
+  # pod's only remote entry point, and getting back in means opening the
+  # case for a serial console.
+  if [ "$SSH_KEY_INSTALLED" = true ]; then
+    sed -i "s/#\?PermitRootLogin .*/PermitRootLogin prohibit-password/" /etc/ssh/sshd_config
+    sed -i "s/#\?PasswordAuthentication .*/PasswordAuthentication no/" /etc/ssh/sshd_config
+    sed -i "s/#\?PermitEmptyPasswords .*/PermitEmptyPasswords no/" /etc/ssh/sshd_config
+  else
+    echo "No SSH key installed — leaving password authentication as-is." >&2
+    echo "Re-run the installer with a valid public key to harden sshd." >&2
   fi
 
   # Validate SSH config before restarting
@@ -1748,11 +1775,17 @@ if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]]; 
     exit 1
   fi
 
+  if [ "$SSH_KEY_INSTALLED" = true ]; then
+    SSH_RESULT="SSH configured on port 8822 (keys only)"
+  else
+    SSH_RESULT="SSH configured on port 8822 (existing auth methods unchanged)"
+  fi
+
   # Restart SSH (try both service names for compatibility)
   if systemctl restart sshd 2>/dev/null; then
-    echo "SSH configured on port 8822 (keys only)"
+    echo "$SSH_RESULT"
   elif systemctl restart ssh 2>/dev/null; then
-    echo "SSH configured on port 8822 (keys only)"
+    echo "$SSH_RESULT"
   else
     echo "Error: Failed to restart SSH service" >&2
     cp /etc/ssh/sshd_config.backup /etc/ssh/sshd_config
--- a/scripts/install
+++ b/scripts/install
@@ -1733,7 +1733,7 @@ if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]]; 
   # Port is safe to set before we know whether a key will land — it doesn't
   # change who can authenticate. The rest of the hardening waits until the
   # key is verifiably in place.
-  sed -i "s/#\?Port .*/Port 8822/" /etc/ssh/sshd_config
+  set_sshd_directive Port 8822
 
   # Repair installs from before the authorized_keys path fix, where the key
   # went to /root/.ssh and sshd (root home = /home/root) never read it.
@@ -1745,14 +1745,25 @@ if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]]; 
 
   SSH_KEY_INSTALLED=false
   if [ -n "$SSH_KEY" ]; then
-    # Validate key format
-    if echo "$SSH_KEY" | grep -qE '^(ssh-rsa|ssh-ed25519|ecdsa-sha2-nistp[0-9]+) [A-Za-z0-9+/=]+'; then
-      AUTHORIZED_KEYS="$(install_authorized_key root "$SSH_KEY")"
-      echo "SSH key added to $AUTHORIZED_KEYS"
-      SSH_KEY_INSTALLED=true
-    else
-      echo "Warning: Invalid SSH key format, skipping"
-    fi
+    # Parse the key with ssh-keygen rather than pattern-matching it. A key
+    # that merely looks well-formed but sshd can't load would leave us
+    # disabling password auth against an unusable credential.
+    KEY_STATUS=0
+    validate_public_key "$SSH_KEY" || KEY_STATUS=$?
+    case "$KEY_STATUS" in
+      0)
+        if AUTHORIZED_KEYS="$(install_authorized_key root "$SSH_KEY")"; then
+          echo "SSH key added to $AUTHORIZED_KEYS"
+          SSH_KEY_INSTALLED=true
+        fi
+        ;;
+      1)
+        echo "Warning: ssh-keygen could not parse that public key, skipping" >&2
+        ;;
+      *)
+        echo "Warning: ssh-keygen is unavailable — cannot verify the key, skipping" >&2
+        ;;
+    esac
   fi
 
   # Only disable password auth once root has a key sshd can actually read.
@@ -1760,9 +1771,9 @@ if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]]; 
   # pod's only remote entry point, and getting back in means opening the
   # case for a serial console.
   if [ "$SSH_KEY_INSTALLED" = true ]; then
-    sed -i "s/#\?PermitRootLogin .*/PermitRootLogin prohibit-password/" /etc/ssh/sshd_config
-    sed -i "s/#\?PasswordAuthentication .*/PasswordAuthentication no/" /etc/ssh/sshd_config
-    sed -i "s/#\?PermitEmptyPasswords .*/PermitEmptyPasswords no/" /etc/ssh/sshd_config
+    set_sshd_directive PermitRootLogin prohibit-password
+    set_sshd_directive PasswordAuthentication no
+    set_sshd_directive PermitEmptyPasswords no
   else
     echo "No SSH key installed — leaving password authentication as-is." >&2
     echo "Re-run the installer with a valid public key to harden sshd." >&2
@@ -1774,6 +1785,26 @@ if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]]; 
     cp /etc/ssh/sshd_config.backup /etc/ssh/sshd_config
     exit 1
   fi
+
+  # A directive can be syntactically valid and still not take effect — an
+  # Include or a later duplicate can override it. Confirm what sshd will
+  # actually do before we restart it into that config.
+  SSH_EXPECTED="port=8822"
+  if [ "$SSH_KEY_INSTALLED" = true ]; then
+    SSH_EXPECTED="$SSH_EXPECTED permitrootlogin=prohibit-password"
+    SSH_EXPECTED="$SSH_EXPECTED passwordauthentication=no permitemptypasswords=no"
+  fi
+  for expectation in $SSH_EXPECTED; do
+    directive="${expectation%%=*}"
+    want="${expectation#*=}"
+    got="$(sshd_effective_value "$directive")"
+    if [ "$got" != "$want" ]; then
+      echo "Error: sshd reports $directive=${got:-<unset>}, expected $want." >&2
+      echo "Restoring backup and leaving sshd untouched..." >&2
+      cp /etc/ssh/sshd_config.backup /etc/ssh/sshd_config
+      exit 1
+    fi
+  done
 
   if [ "$SSH_KEY_INSTALLED" = true ]; then
     SSH_RESULT="SSH configured on port 8822 (keys only)"
--- a/scripts/lib/ssh-helpers
+++ b/scripts/lib/ssh-helpers
@@ -77,11 +77,15 @@ resolve_authorized_keys_path() {
     return 0
   fi
 
-  # %% must expand last so a literal %%h stays literal.
+  # Protect a literal %% before the token passes, not after: in `%%h` the %h
+  # pass matches the second % plus the h, so expanding %% last would still
+  # yield `%<home>`. The sentinel is a byte sshd_config cannot contain.
+  local sentinel=$'\x01'
+  keyfile="${keyfile//%%/$sentinel}"
   keyfile="${keyfile//%h/$home}"
   keyfile="${keyfile//%U/$uid}"
   keyfile="${keyfile//%u/$user}"
-  keyfile="${keyfile//%%/%}"
+  keyfile="${keyfile//$sentinel/%}"
 
   case "$keyfile" in
     /*) printf '%s\n' "$keyfile" ;;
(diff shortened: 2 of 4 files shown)


scripts/install
```diff
@@ -1713,46 +1715,77 @@ fi
...
   # Backup sshd_config
   if [ ! -f /etc/ssh/sshd_config.backup ]; then
     cp /etc/ssh/sshd_config /etc/ssh/sshd_config.backup
   fi
 
-  # Update SSH configuration
+  # Port is safe to set before we know whether a key will land — it doesn't
+  # change who can authenticate. The rest of the hardening waits until the
+  # key is verifiably in place.
   sed -i "s/#\?Port .*/Port 8822/" /etc/ssh/sshd_config
```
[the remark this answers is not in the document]
jasonlarue: Fixed in 62e56fd. Confirmed: the substitutions only rewrote directives that were already present, so a config that never mentioned `PasswordAuthentication` kept sshd's default of `yes` while the installer reported it had hardened.

Replaced them with `set_sshd_directive`, which rewrites the first occurrence, drops later duplicates so nothing downstream re-overrides it, appends the directive when absent, and inserts above the first `Match` rather than after it (a bare append lands *inside* the last Match block). It also stays out of Match blocks entirely, since a directive there is scoped to that block and overwriting it would change semantics this PR has no business touching.

Written in awk, not sed: `sed -i -E` is not portable — BSD sed takes the next argument as a backup suffix, so the sed version worked on the pod and silently no-opped on macOS. That was hiding two test failures.

Also added the `sshd -T` assertion you asked for: after the edits, the installer re-reads the effective `port`/`permitrootlogin`/`passwordauthentication`/`permitemptypasswords` values and restores the backup and exits if any of them is not what it just set.

scripts/install
```diff
@@ -1713,46 +1715,77 @@ fi
...
     # Validate key format
     if echo "$SSH_KEY" | grep -qE '^(ssh-rsa|ssh-ed25519|ecdsa-sha2-nistp[0-9]+) [A-Za-z0-9+/=]+'; then
-      mkdir -p /root/.ssh
-      echo "$SSH_KEY" >> /root/.ssh/authorized_keys
-      chmod 700 /root/.ssh
-      chmod 600 /root/.ssh/authorized_keys
-      echo "SSH key added"
+      AUTHORIZED_KEYS="$(install_authorized_key root "$SSH_KEY")"
+      echo "SSH key added to $AUTHORIZED_KEYS"
+      SSH_KEY_INSTALLED=true
```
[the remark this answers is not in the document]
jasonlarue: Fixed in 62e56fd. You're right that the regex accepted `ssh-ed25519 A` — sshd ignores a line like that, so the installer would have disabled password auth against a credential that can never authenticate.

Now `validate_public_key` parses the key with `ssh-keygen -l` and returns three states: 0 valid, 1 rejected, 2 unable to tell. The installer only sets `SSH_KEY_INSTALLED=true` on 0, and treats 2 (no `ssh-keygen` on PATH) as a reason to leave password auth alone rather than harden blind.

One deviation from the suggestion: it writes to a `mktemp` file instead of `ssh-keygen -l -f /dev/stdin`. Same check, but `/dev/stdin` handling varies across the busybox-ish userlands these pods ship, and this is the code path where being wrong means a locked-out pod.

scripts/lib/ssh-helpers
```diff
@@ -0,0 +1,115 @@
...
+  [ -n "$keyfile" ] || keyfile=".ssh/authorized_keys"
+
+  # The directive takes a list; sshd reads every entry but writing the first
+  # is enough to authenticate.
+  keyfile="${keyfile%%[[:space:]]*}"
+
+  # %% must expand last so a literal %%h stays literal.
+  keyfile="${keyfile//%h/$home}"
+  keyfile="${keyfile//%u/$user}"
+  keyfile="${keyfile//%%/%}"
```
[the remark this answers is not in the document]
jasonlarue: Fixed in 62e56fd — all three parts.

`sshd -T` is now called as `-C user=root,host=localhost,addr=127.0.0.1`, so a `Match User root` override is evaluated the way it would be for a real root login. Older sshd rejects a partial connection spec, so a plain `sshd -T` retry sits behind it, and the raw-config awk fallback is unchanged and still independent (it reports the global value — noted in a comment, since it cannot evaluate Match).

`AuthorizedKeysFile none` now propagates as the literal `none` instead of being expanded into a path. `install_authorized_key` refuses it with a message pointing at sshd_config, `migrate_stranded_authorized_keys` no-ops, and the installer leaves password auth untouched because `SSH_KEY_INSTALLED` stays false.

`%U` expands from `id -u`, ordered before `%u` so the two do not collide, with `%%` still last.

Regression coverage added for all three: `applies a Match User root override, which only -C surfaces`, `falls back to a plain sshd -T when the installed sshd rejects -C`, `refuses to invent a path when AuthorizedKeysFile is none`, `is a no-op when key files are disabled entirely`, and `expands %U to the target uid`.

scripts/lib/ssh-helpers
```diff
@@ -0,0 +1,115 @@
...
+  local migrated=0 line
+  while IFS= read -r line; do
+    [ -n "$line" ] || continue
+    case "$line" in \#*) continue ;; esac
+    if ! grep -qxF "$line" "$keyfile" 2>/dev/null; then
+      mkdir -p "$(dirname "$keyfile")"
+      printf '%s\n' "$line" >> "$keyfile"
+      migrated=$((migrated + 1))
+    fi
+  done < "$stranded"
```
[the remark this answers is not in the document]
jasonlarue: Fixed in 62e56fd — now `while IFS= read -r line || [ -n "$line" ]`. Worth fixing precisely because this reads a file an earlier installer wrote and a user may since have hand-edited; a missing trailing newline would have silently dropped the key being recovered, which is the one key that matters here.

Covered by `recovers a final key with no trailing newline`.

scripts/lib/ssh-helpers
```diff
@@ -0,0 +1,245 @@
...
+    printf '%s\n' "$key" >> "$keyfile"
+  fi
+
+  # StrictModes rejects a key file (or its directory) that is group- or
+  # world-writable, or owned by anyone but the user or root.
+  chown "$user" "$keydir" "$keyfile" 2>/dev/null || true
+  chmod 700 "$keydir"
+  chmod 600 "$keyfile"
+
+  printf '%s\n' "$keyfile"
```
ng, line 165: **Blocking:** this function can report success without a key on disk, which re-creates the exact lockout this PR exists to prevent.

The call site tests the result (`if AUTHORIZED_KEYS="$(install_authorized_key ...)"`), so bash suppresses `set -e` inside the entire function. A failed `mkdir`, a failed append, or a failed `chmod` is silently ignored and the final `printf` still returns 0 — after which the installer sets `SSH_KEY_INSTALLED=true` and disables password auth against a key that was never written.

Repro: stub root's home to a `chmod 555` directory → the function prints the keyfile path and returns 0; nothing exists on disk. Disk pressure makes this real on pods (see the PR #659 pruner incident), and an ENOSPC append can land a *partial* key line that sshd silently ignores. The `sshd -T` verification loop can't catch either case — it checks directives, not the key file.

```suggestion
  keydir="$(dirname "$keyfile")"
  mkdir -p "$keydir" || return 1
  if ! grep -qxF "$key" "$keyfile" 2>/dev/null; then
    printf '%s\n' "$key" >> "$keyfile"
  fi

  # StrictModes rejects a key file (or its directory) that is group- or
  # world-writable, or owned by anyone but the user or root.
  chown "$user" "$keydir" "$keyfile" 2>/dev/null || true
  chmod 700 "$keydir" || return 1
  chmod 600 "$keyfile" || return 1

  # The caller disables password auth on our word — don't take the write on
  # faith. An ENOSPC append can also land a partial line sshd will silently
  # ignore, so require the exact key line to be present.
  grep -qxF "$key" "$keyfile" 2>/dev/null || return 1

  printf '%s\n' "$keyfile"
```

Pinning test (needs `chmodSync` added to the `node:fs` import):

```ts
test('fails rather than reporting success when the key cannot be written', () => {
  const home = join(dir, 'home', 'root')
  mkdirSync(home, { recursive: true })
  chmodSync(home, 0o555)

  const snippet = `install_authorized_key root ${JSON.stringify(realKey)} 2>/dev/null || echo REFUSED`
  const output = runBash(snippet, { rootHome: home })
  chmodSync(home, 0o755)

  expect(output).toBe('REFUSED')
})
```

scripts/lib/ssh-helpers
```diff
@@ -0,0 +1,245 @@
...
+  if [ "$keyfile" = "none" ]; then
+    printf 'none\n'
+    return 0
+  fi
+
+  # %% must expand last so a literal %%h stays literal.
+  keyfile="${keyfile//%h/$home}"
+  keyfile="${keyfile//%U/$uid}"
+  keyfile="${keyfile//%u/$user}"
+  keyfile="${keyfile//%%/%}"
```
ng: The comment promises something substitution order can't deliver: in `%%h`, the `%h` pass matches the second `%` plus the `h` first, so `%%h/keys` comes out as `%/home/root/keys` (verified), not literal `%h/keys`. Ordering `%%` last doesn't protect it — it has to be protected *first*:

```suggestion
  # Protect literal %% before the token passes — otherwise the %h pass eats
  # the second % of `%%h` and expands it. The sentinel is a byte sshd_config
  # cannot contain.
  local sentinel=$'\x01'
  keyfile="${keyfile//%%/$sentinel}"
  keyfile="${keyfile//%h/$home}"
  keyfile="${keyfile//%U/$uid}"
  keyfile="${keyfile//%u/$user}"
  keyfile="${keyfile//$sentinel/%}"
```

A literal `%` in `AuthorizedKeysFile` is vanishingly rare, so downgrading this to a comment-only fix would also be fine — just don't ship a comment asserting a guarantee the code doesn't provide.

scripts/lib/ssh-helpers
```diff
@@ -0,0 +1,245 @@
...
+      if (!in_match && tolower(probe) ~ /^match[ \t]/) {
+        in_match = 1
+        if (!done) { print key " " value; done = 1 }
+      }
+
+      if (!in_match) {
+        candidate = probe
+        sub(/^#/, "", candidate)
+        split(candidate, parts, /[ \t]+/)
+        if (tolower(parts[1]) == tolower(key) && candidate ~ /^[^ \t]+[ \t]/) {
```
ng: Non-blocking: sshd_config also accepts `Key=value` syntax (`PasswordAuthentication=no` is valid), which this whitespace split doesn't recognise — a stale `=`-form directive earlier in the file would survive and win over the appended one (sshd takes the first occurrence). The `sshd -T` verification downstream catches the mismatch and aborts fail-closed, so it's safe — just a mysterious failure if ever hit. Worth a comment here.

scripts/install
```diff
@@ -1713,32 +1715,68 @@ fi
 if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]]; then
   echo "Configuring SSH..."
 
+  if [ ! -f "$INSTALL_DIR/scripts/lib/ssh-helpers" ]; then
+    # Without the helpers we can't tell where sshd reads authorized_keys
+    # from, and hardening on a guess locks the pod out for good. Leave sshd
+    # exactly as it is.
+    echo "Error: $INSTALL_DIR/scripts/lib/ssh-helpers not found — skipping SSH setup." >&2
```
ng: This says "skipping SSH setup" but the `exit 1` aborts the whole installer — IP detection, the service-start wait, and the completion banner never run. The abort is the right call (fail-closed); the message should own it:

```suggestion
    echo "Error: $INSTALL_DIR/scripts/lib/ssh-helpers not found — aborting install." >&2
```

scripts/install
```diff
@@ -1713,32 +1715,68 @@ fi
...
+    echo "sshd was left unchanged. Re-run the installer once the source tree is complete." >&2
+    exit 1
+  fi
+  source "$INSTALL_DIR/scripts/lib/ssh-helpers"
+
   # Backup sshd_config
   if [ ! -f /etc/ssh/sshd_config.backup ]; then
     cp /etc/ssh/sshd_config /etc/ssh/sshd_config.backup
   fi
```
ng: `sshd_config.backup` is frozen at the *first ever* installer run (only created if absent). This PR adds a new restore path — the verification loop below — that copies it over the live config; on a re-run that reverts the pod to the stock config (port 22, password auth on), which is neither "untouched" nor what any runbook expects. Keep the historical artifact, but restore from a fresh per-run snapshot:

```suggestion
  # Keep the first-ever config as a historical artifact, but restore from a
  # fresh per-run snapshot below — .backup is frozen at the first run, so
  # restoring it on a re-run reverts to the stock config (port 22, password
  # auth on) rather than to the state this run started from.
  if [ ! -f /etc/ssh/sshd_config.backup ]; then
    cp /etc/ssh/sshd_config /etc/ssh/sshd_config.backup
  fi
  cp -f /etc/ssh/sshd_config /etc/ssh/sshd_config.pre-install
```

scripts/install
```diff
@@ -1748,11 +1786,37 @@ if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]];
...
+    SSH_EXPECTED="$SSH_EXPECTED passwordauthentication=no permitemptypasswords=no"
+  fi
+  for expectation in $SSH_EXPECTED; do
+    directive="${expectation%%=*}"
+    want="${expectation#*=}"
+    got="$(sshd_effective_value "$directive")"
+    if [ "$got" != "$want" ]; then
+      echo "Error: sshd reports $directive=${got:-<unset>}, expected $want." >&2
+      echo "Restoring backup and leaving sshd untouched..." >&2
+      cp /etc/ssh/sshd_config.backup /etc/ssh/sshd_config
```
ng: With the per-run snapshot above, restore from it here — that also makes the "leaving sshd untouched" message accurate, which restoring the first-run `.backup` does not:

```suggestion
      cp /etc/ssh/sshd_config.pre-install /etc/ssh/sshd_config
```

The same change applies at the pre-existing `sshd -t` failure path a few lines up (outside this diff) and the restart-failure path below.

scripts/install
```diff
@@ -1748,11 +1786,37 @@ if [ "${SKIP_SSH:-}" != true ] && [ -t 0 ] && [[ "${SETUP_SSH:-}" =~ ^[Yy]$ ]];
...
   # Restart SSH (try both service names for compatibility)
   if systemctl restart sshd 2>/dev/null; then
-    echo "SSH configured on port 8822 (keys only)"
+    echo "$SSH_RESULT"
   elif systemctl restart ssh 2>/dev/null; then
-    echo "SSH configured on port 8822 (keys only)"
+    echo "$SSH_RESULT"
   else
     echo "Error: Failed to restart SSH service" >&2
     cp /etc/ssh/sshd_config.backup /etc/ssh/sshd_config
```
ng: Per-run snapshot here too:

```suggestion
    cp /etc/ssh/sshd_config.pre-install /etc/ssh/sshd_config
```

ng (changes requested):
Reviewed in a local worktree: branch checked out, full new test suite run (28/28 pass), and the two main claims below verified by actually executing the helper code.

The approach is right and most of it is carefully built — the resolution order (`sshd -T -C` → `sshd -T` → raw config → compiled-in default), hardening ordered strictly after a verified key install, and the effective-config check before restart are all sound, and the tests pin real behavior. One confirmed bug re-opens the exact lockout class this PR closes — flagged inline with an applyable fix. The rest are minors/nits.

**Confirmed by running the code:**
- `install_authorized_key` reports success while writing nothing when the target directory isn't writable (repro: root home on a `chmod 555` dir → prints the path, returns 0, no file on disk). See inline — this is the one blocking item.
- The `%%` token comment in `resolve_authorized_keys_path` promises an ordering guarantee bash substitution can't deliver: `%%h/keys` expands to `%/home/root/keys`, not literal `%h/keys`.

**Checked and sound:** every `|| true` guard against `set -euo pipefail`; migration dedup (`grep -qxF`) also implicitly covering a symlinked `/root` → `/home/root`; `set_sshd_directive`'s Match-block / prose-comment / duplicate handling against every edge I could construct; `cat > config` preserving inode and modes; single-line `read -r` keeping a pasted private key out of `validate_public_key`.

**Vs. CodeRabbit:** its four findings (missing-directive hardening gap, lookalike-key regex, `Match`/`-C` evaluation, trailing-newline migration loss) were all addressed in 62e56fd and confirmed by the bot. Nothing below overlaps — these are new findings on top.

Nit: the PR body says 12 tests — the file has 28.

scripts/lib/ssh-helpers
jasonlarue: Fixed in c594a20 — applied your suggestion as written. Thanks for running it down; the `set -e` suppression from the `if` at the call site is exactly the kind of thing that reads fine and is silently fatal.

I reproduced both sides before and after to confirm the test is a real pin, not just a passing assertion: against the pre-fix function body a `chmod 555` home makes it print `/…/home/root/.ssh/authorized_keys` and return 0 with no file on disk; after the fix the same input returns non-zero. Test added as `fails rather than reporting success when the key cannot be written`.

The read-back `grep -qxF` earns its place independently of the guards — a partial ENOSPC append leaves a truncated line that sshd ignores, and no exit status anywhere reports that.

scripts/lib/ssh-helpers
jasonlarue: Fixed in c594a20 with the sentinel, not the comment-only downgrade. You are right that the comment asserted a guarantee the code did not have: `%h` matches the second `%` of `%%h` first, so ordering `%%` last was never protection.

Took the code fix over the comment fix because the sentinel is three lines and removes the class of bug rather than documenting it, and because this function feeds a path that gets written to and then hardened against — a wrong path here is the failure mode the whole PR is about. Pinned by `keeps a literal %% from being eaten by the %h pass` (`/etc/ssh/%%h/authorized_keys` → `/etc/ssh/%h/authorized_keys`).

scripts/lib/ssh-helpers
jasonlarue: Handled the syntax rather than just commenting it, in c594a20 — it turned out cheaper than the comment would have been to justify.

The awk now extracts the directive name as the leading run of non-space, non-`=` characters and accepts any of `Key value`, `Key=value`, or `Key = value` as the separator. Prose comments still fall out for free: `# Port forwarding …` leaves the extracted name empty (the space follows the `#`), so it can never match.

Agreed the `sshd -T` check downstream made this safe rather than dangerous — but "safe" here meant a fail-closed abort with a confusing message, and this removes the confusing message. Covered by `replaces the Key=value form sshd also accepts` and `replaces a commented Key=value directive`.

scripts/install
jasonlarue: Fixed in c594a20 — now reads "aborting install". You are right that the branch exits 1 and everything downstream is skipped; the message was describing an intent the code does not have.

scripts/install
jasonlarue: Fixed in c594a20 — applied as suggested. This one I had not considered at all: I added a new restore path in 62e56fd without noticing that `.backup` is created only when absent, so on any pod that has run the installer before, my verification loop would have rolled sshd back to the stock config — port 22, password auth on — and called it "leaving sshd untouched". That is a worse outcome than the failure it was reacting to.

`.backup` keeps its historical role; `/etc/ssh/sshd_config.pre-install` is refreshed at the top of every run and is what all restores now use.

scripts/install
jasonlarue: Fixed in c594a20. Updated all three restore sites, including the pre-existing `sshd -t` failure path above and the restart-failure path below as you flagged — leaving those on the frozen `.backup` would have kept the exact bug for two of the three ways this block can fail.

scripts/install
jasonlarue: Fixed in c594a20 — covered by the same change.

jasonlarue:
@ng — all seven addressed in c594a20, replies inline on each thread. Summary:

**Blocking:** `install_authorized_key` now checks every step and reads the exact key line back before printing the path. I rebuilt the pre-fix function body and confirmed your repro (prints the path, returns 0, nothing on disk) and that it now returns non-zero, so the new test is a real pin rather than a passing assertion.

**Per-run snapshot:** this was the one I am most glad you caught — 62e56fd added a restore path without noticing `.backup` is frozen at the first-ever run, so my own verification loop would have rolled any re-run back to the stock config and reported it as "leaving sshd untouched". All three restore sites now use `/etc/ssh/sshd_config.pre-install`.

**Took the code fix over the comment-only option twice:** the `%%` sentinel, and `Key=value` recognition in `set_sshd_directive`. Both were small enough that removing the class of bug beat documenting it, and both are pinned.

Also fixed the message on the helper-missing branch, and the 12 → 32 test count in the PR description.

32/32 helper tests pass; full suite 3340 passed / 1 skipped across 137 files; `pnpm tsc` and `pnpm lint` clean.

One note outside the diff: GitHub Actions still has not run on this PR — zero workflow runs for any head SHA, which is the fork-PR approval gate rather than anything in the branch. It needs an "Approve and run workflows" click whenever you next look.

scripts/lib/ssh-helpers
```diff
@@ -0,0 +1,269 @@
...
+  local sshd_config="${2:-/etc/ssh/sshd_config}"
+  local home uid keyfile
+
+  # The `|| true` guards matter: scripts/install runs under `set -euo
+  # pipefail`, where a missing getent or a non-zero sshd would otherwise
+  # abort the installer mid-hardening.
+  home="$(getent passwd "$user" 2>/dev/null | cut -d: -f6 || true)"
+  [ -n "$home" ] || home="/root"
+  uid="$(id -u "$user" 2>/dev/null || true)"
+  [ -n "$uid" ] || uid="0"
```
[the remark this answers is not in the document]
jasonlarue: Fixed in edd26a0. Agreed this is the sharpest of the three — the `/root` fallback is the original bug's exact shape, just reached by a different route: a key installed and read back successfully in a file sshd never opens, followed by `PasswordAuthentication no`.

Resolution now tries `getent`, then reads `/etc/passwd` directly (getent is glibc's; the fallback covers an image that ships without it, which is the realistic reason for it to be unavailable), and fails if neither yields a home. `install_authorized_key` propagates that with `|| return 1`, and the test that expected `/root/.ssh/authorized_keys` now expects a refusal.

One deviation: `migrate_stranded_authorized_keys` skips with a warning rather than propagating. It is called bare under `set -e`, so returning non-zero would abort an otherwise-complete install at the very last step — and skipping it is not the dangerous branch, because `install_authorized_key` fails on the same condition and that is what actually keeps the hardening from running. Pinned by `skips rather than aborting the installer when the path is unresolvable`.

scripts/lib/ssh-helpers
```diff
@@ -0,0 +1,269 @@
...
+        }
+      }
+
+      print $0
+    }
+    END { if (!done) print key " " value }
+  ' "$sshd_config" > "$tmp"; then
+    cat "$tmp" > "$sshd_config"
+    rm -f "$tmp"
+    return 0
```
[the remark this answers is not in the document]
jasonlarue: Fixed in edd26a0. Both halves were real: `cat > "$sshd_config"` truncates the live file before writing, and its exit status was unchecked.

Now: `readlink -f` first so a symlinked sshd_config is replaced at the real file rather than clobbering the link; `mktemp` beside the target so the rename is same-filesystem and atomic; `cp -p` to seed the temp with the original mode and owner; `mv -f` only after awk succeeds; failures cleaned up and propagated.

Worth flagging explicitly for @ng, who checked the previous version specifically for inode and mode preservation: **the rename changes the inode.** Mode and ownership are carried across (pinned by `preserves the config file mode across the rewrite`), but anything holding the old inode — a hardlink, a bind mount of the file itself — would no longer see updates. On these pods `/etc/ssh/sshd_config` is a regular file, so I took atomicity over inode stability; happy to revert if that trade is wrong for a reason I cannot see from here.

Write-failure regression added as `reports failure and leaves the config intact when it cannot be replaced` — read-only parent directory, asserts non-zero *and* that the original config is unmodified.

scripts/lib/ssh-helpers
```diff
@@ -0,0 +1,269 @@
...
+# Print sshd's effective value for directive $1, so the installer can confirm
+# what it asked for is what sshd will actually do before restarting it.
+sshd_effective_value() {
+  local name="$1" sshd_config="${2:-/etc/ssh/sshd_config}"
+  local lowered
+  lowered="$(printf '%s' "$name" | tr '[:upper:]' '[:lower:]')"
+
+  sshd -T -f "$sshd_config" 2>/dev/null |
+    awk -v want="$lowered" 'tolower($1) == want { print $2; exit }' \
+    || true
```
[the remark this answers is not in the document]
jasonlarue: Fixed in edd26a0. The gap is real, though the direction is worth naming precisely: a `Match User root` block re-enabling `PasswordAuthentication yes` means the installer reports a hardened pod that still accepts a root password. That is a weaker guarantee than advertised rather than a lockout — but "we told you it was keys-only" is exactly the claim this PR exists to make trustworthy.

`sshd_effective_value` takes an optional user now, and the installer verifies `PermitRootLogin`, `PasswordAuthentication` and `PermitEmptyPasswords` with `-C user=root,host=localhost,addr=127.0.0.1`. `Port` stays global, since Match cannot set it.

On failing closed when `-C` is unusable: I degraded to the global reading instead, mirroring the fallback already in `resolve_authorized_keys_path`. Failing closed there would abort the install on any sshd too old to accept a partial connection spec — turning an unsupported flag into a hard failure on exactly the older firmware this script exists to support. Both paths are pinned: `sees a Match User root override that the global reading misses` and `degrades to the global reading when sshd rejects -C`.

Both README updates applied — the hardening section now says verification happens in root context, and the manual instruction is the `-C` form with a note on why a plain `sshd -T` is not enough.

jasonlarue:
CodeRabbit's three findings on c594a20 are addressed in edd26a0; replies inline on each thread.

- **`/root` home fallback (critical)** — genuinely the original bug reached by another route, so the guess is gone: `getent`, then `/etc/passwd`, then fail. `install_authorized_key` propagates; `migrate_stranded_authorized_keys` skips instead, because it runs bare under `set -e` and its failure is not the dangerous one.
- **Non-atomic config replace** — temp file beside the target, `cp -p` to carry mode and owner, `mv` after awk succeeds, symlinks resolved first, every step checked. @ng: this changes the inode, which you had checked and approved as preserved — details and my reasoning in the thread, easy to revert if that trade is wrong.
- **Match-context verification** — the auth directives are now read with `sshd -T -C user=root,…`; `Port` stays global. I degraded to the global reading when `-C` is unsupported rather than failing closed, so an older sshd does not turn into a failed install; reasoning in the thread.

Tests 32 → 39, including a write-failure regression that asserts the config is left unmodified. Full suite 3347 passed / 1 skipped across 137 files; `pnpm tsc` and `pnpm lint` clean.

@ng — your seven from the previous round are still addressed as described in c594a20; nothing here reverted any of them. The fork-PR workflow approval is still outstanding, so CI has yet to run on any commit in this branch.

jasonlarue:
Both items from the review of edd26a0 are addressed in 863ee0b. Replying here since both landed in the review body rather than on inline threads.

**`Include` ordering (Major) — valid, fixed.** sshd splices an included file in where the `Include` appears and keeps the first value it obtains per keyword, so a directive written below one loses to whatever the Include sets. `set_sshd_directive` only treated `Match` as a boundary, so on an Include-based config it would rewrite the main file, report success, and then the effective-value check would fail and roll the install back. Fail-closed, so not a lockout — but an abort that no amount of retrying fixes, which is a bad place to leave someone mid-install.

`Include` is now a boundary alongside `Match`: the directive is emitted above the first one, existing occurrences that already precede it are still rewritten in place, and stale duplicates below it are still dropped. I took the first of the two options offered (set it before the first applicable Include) rather than resolving and rewriting the included source — same guarantee, and it keeps this function to one file and one pass, which matters given it edits the file that gates all remote access.

I checked this against the committed version rather than assuming: on `Include …` + `X11Forwarding yes`, edd26a0 appends `PasswordAuthentication no` at the bottom (where an Include-set value beats it); 863ee0b emits it at the top. Three tests added — above-Include insertion, existing directive already above the Include left in place, and stale duplicate below it dropped.

**Root-privilege test flakiness (nit) — valid, and it applied to two tests, not one.** The `install_authorized_key` failure test had the same problem; it was @ng's original repro and I had carried the 0555 approach into the `set_sshd_directive` one. As uid 0 the mode bits are ignored, the command succeeds, and the assertion fails for the wrong reason.

Each now has a stub-driven twin that runs anywhere — `mv() { return 1; }` for the replacement path, `mkdir() { return 1; }` for the key write — and the filesystem-level versions are kept behind `test.skipIf(isRoot)`. I did not want to lose them outright: they exercise real filesystem behaviour rather than a shell function shadowing a builtin, and that is the check @ng actually reproduced by hand.

Tests 39 → 44. Full suite 3352 passed / 1 skipped across 137 files; `pnpm tsc` and `pnpm lint` clean.

scripts/lib/ssh-helpers
```diff
@@ -0,0 +1,329 @@
...
+  local migrated=0 line
+  # `|| [ -n "$line" ]` keeps the last line when the file has no trailing
+  # newline — a hand-edited authorized_keys often doesn't.
+  while IFS= read -r line || [ -n "$line" ]; do
+    [ -n "$line" ] || continue
+    case "$line" in \#*) continue ;; esac
+    if ! grep -qxF "$line" "$keyfile" 2>/dev/null; then
+      mkdir -p "$(dirname "$keyfile")"
+      printf '%s\n' "$line" >> "$keyfile"
+      migrated=$((migrated + 1))
```
ng: Nit (optional, non-blocking): this function is documented as best-effort — *"an unresolvable home is a skip, not a failure … returning non-zero would abort an otherwise-complete install."* But it's called bare at `scripts/install:1744` under `set -euo pipefail`, and these two mutations are unguarded — an ENOSPC on the append (or a `mkdir` failure) would trip `set -e` and abort the whole install right before the completion banner. Guarding each keeps the skip-not-fail contract:

```suggestion
      mkdir -p "$(dirname "$keyfile")" || continue
      printf '%s\n' "$line" >> "$keyfile" || continue
      migrated=$((migrated + 1))
```

scripts/lib/ssh-helpers
```diff
@@ -0,0 +1,329 @@
...
+    fi
+  done < "$stranded"
+
+  if [ "$migrated" -gt 0 ]; then
+    echo "Recovered $migrated SSH key(s) from $stranded -> $keyfile"
+    chown "$user" "$(dirname "$keyfile")" "$keyfile" 2>/dev/null || true
+    chmod 700 "$(dirname "$keyfile")"
+    chmod 600 "$keyfile"
```
ng: Same best-effort nit: match the adjacent `chown … 2>/dev/null || true` so a `chmod` failure here can't abort the install either:

```suggestion
    chmod 700 "$(dirname "$keyfile")" 2>/dev/null || true
    chmod 600 "$keyfile" 2>/dev/null || true
```

ng (approved):
Re-reviewed at 62f5eaca. Both items from my earlier CHANGES_REQUESTED are resolved:

- `install_authorized_key` now verifies the key is actually on disk before reporting success (`grep -qxF "$key" "$keyfile" || return 1`, plus `|| return 1` on `mkdir`/`chmod`) — the write-nothing / report-success lockout path is closed. Pinned by the new "fails rather than reporting success when the key cannot be written" and unwritable-home tests.
- The `%%` ordering is now handled with a `$'\x01'` sentinel; new test `keeps a literal %% from being eaten by the %h pass` covers it.

The resolution order, hardening-after-verified-key ordering, and the effective-config (`sshd -T`, root context) check before restart all hold up. **Approving.**

Two optional, non-blocking nits inline on `migrate_stranded_authorized_keys`: the function documents itself as best-effort ("a skip, not a failure … returning non-zero would abort an otherwise-complete install"), but a few of its filesystem mutations are unguarded while it's called bare under `set -euo pipefail`. Safe to apply or ignore.

scripts/lib/ssh-helpers
```diff
@@ -0,0 +1,329 @@
...
+  # are equivalent to sshd, and matching only the first would leave a stale
+  # `=`-form directive in place to win over the one we append (sshd takes
+  # the first occurrence).
+  #
+  # `Include` is a boundary as well as `Match`: sshd splices the included
+  # file in at that point and keeps the first value it obtains, so a directive
+  # set by an early Include beats anything written further down this file.
+  # Emitting ours above the first Include is what makes it the winning value —
+  # otherwise set_sshd_directive reports success and the installer's
+  # effective-value check then fails and rolls the whole thing back.
```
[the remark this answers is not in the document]
ng: Fixed in a703e40. The boundary keyword is now extracted with the same `sub(/[ \t=].*$/, "", …)` rule the directive-name match already used, so `Include=/path` and `Match=User x` register as boundaries exactly like their space-separated forms. Without it, our directive appended below an `Include=…` and lost to the value sshd had already taken from the included file. Regression tests added: `treats the Include=path form as a boundary too` and `does not rewrite a directive inside a Match=criteria block`.

ng:
Pushed a703e40 to close out the remaining review threads:

- **`sshd_effective_value` now fails closed (CodeRabbit critical).** It no longer degrades to a plain `sshd -T` when the root-context `-C` query is unavailable. A global reading can't see a `Match User root` block re-enabling password auth for the one account being hardened, so trusting it would let the installer claim a hardened pod it can't verify. It now emits nothing; the installer reads the empty result as a verification failure and restores the pre-install snapshot. This reverses the earlier "degrade to global" note deliberately — it brings the function in line with this file's stated fail-closed contract. The only case that changes is an sshd too old to accept the partial `-C` spec (pre-6.7), which no pod ships; there it now aborts+restores instead of hardening blind. Test and `scripts/README.md` updated.

- **`Include=`/`Match=` equals form recognized as a boundary (CodeRabbit major).** See the inline reply.

- **`migrate_stranded_authorized_keys` keeps its best-effort contract (our review nits).** `mkdir`/append/`chmod` are guarded so a disk-pressure failure skips the recovery rather than tripping `set -e` and aborting an otherwise-complete install.

Gates: `pnpm tsc` clean, `pnpm test` 3354 passed / 1 skipped across 137 files, `bash -n` clean on both scripts.

PR #703, opened 2026-09-08
fix(deploy): require prebuilt updates and stage local builds over SSH

ng:
A branch without a GitHub release previously made `sp-update` replace the installed app and attempt a full dependency install and Next.js build on the Pod. Updates now require a validated prebuilt bundle and report missing releases, API errors, and failed downloads without falling back to source. Staging and rollback use persistent storage, backups must succeed, and symlinked installation paths are resolved before stale files are removed.

From a clone on macOS or Linux, `./scripts/deploy POD_IP [branch]` installs locked dependencies, builds locally with isolated in-memory databases, and uploads a complete archive before invoking `sp-update --archive`. Optional origin branches build in a temporary worktree without changing the user's checkout. The archive retains the standalone runtime and build metadata while excluding environment files, Git metadata, and workstation node_modules. The Pod installs production dependencies for its own architecture.

Validation:
- Deployed commit `9544003` to Pod 88 through the local build/upload/update path. The running version API reports that exact commit; UI, version API, and OpenAPI return HTTP 200. Firmware connection, sensor frames, and 329 scheduled jobs initialized, with zero service restarts after startup. The Pod's `.env` checksum is unchanged.
- Live testing exercised code/database rollback after a failed start and exposed the SSH PATH, build-worker SQLite contention, symlink cleanup, and standalone-runtime issues fixed in this PR.
- 24 targeted deployment/firewall tests pass, along with TypeScript, ESLint, both Drizzle schema checks, Bash syntax, and ShellCheck at warning severity. GitHub CI builds and unit tests pass. ESLint has one existing warning in stryker.config.mjs.
- CodeRabbit reviewed the initial commit; all five findings were addressed. Its follow-up review of the latest changes was rate-limited for 45 minutes.

Limits: the Pod still needs WAN access for production/native dependencies. Rollback restores code/database but does not restore node_modules or module/systemd changes. First-time installations continue to use scripts/install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,10 +50,20 @@ After install, these are available system-wide:
 sp-status         # systemctl status sleepypod.service
 sp-restart        # restart the service
 sp-logs           # journalctl -u sleepypod.service -f
-sp-update         # pull latest, rebuild, migrate, restart (with automatic rollback)
+sp-update         # install latest pre-built release, migrate, restart (with rollback)
 sp-maintenance    # one-shot manual prime / reboot / status
 sp-uninstall      # stop services, remove systemd units, optionally wipe data
 ```
+
+Cloned the repo on your computer? Build locally and deploy to an installed Pod:
+
+```bash
+./scripts/deploy POD_IP              # current checkout
+./scripts/deploy POD_IP fix/my-fix   # branch from your clone's origin
+```
+
+See the [deployment guide](docs/DEPLOYMENT.md) for SSH requirements and fork/PR validation.
+
 
 ---
 
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -26,7 +26,7 @@ flowchart TB
     end
 
     subgraph trigger["Trigger Sources"]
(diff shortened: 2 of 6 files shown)


ng:
@coderabbitai review

ng:
@coderabbitai review — please include the latest fixes from live Pod 88 validation: SSH PATH normalization, in-memory build databases, and resolving symlinked installation paths before cleanup (97e4f5c).

ng:
Addressed all five findings from the initial review: ordering assertions now require marker presence; the curl stub has a bounded argument scan; early deploy failures preserve an empty log; fixture path rewrites assert their inputs; and TMPDIR capacity is checked before stopping the service.

Live Pod 88 testing also uncovered and fixed SSH PATH selection, parallel build workers contending over local SQLite files, stale standalone code surviving cleanup through a symlinked installation path, and the need to retain the standalone runtime while excluding workstation node_modules. The failed-start path restored the prior code/database, and the previous service is running. A fresh deployment of the corrected archive is next.

@coderabbitai review

PR #705, opened 2026-09-09
fix(deploy): support explicit GitHub forks for branch builds

ng:
Deploying `fix/pod3-frozen-heartbeat` from an upstream clone fails because that branch exists in `Kovbo/core`, while the local deploy script fetches only from `origin`.

Add `./scripts/deploy --repo Kovbo/core POD_IP fix/pod3-frozen-heartbeat` to fetch the fork branch into the existing temporary worktree without changing remotes, the current branch, or local edits. `SLEEPYPOD_GITHUB_REPO` selects the same repository; `--repo` takes precedence. Explicit repositories require a branch. With no selector, existing origin/current-checkout behavior is preserved.

Validation: 30 targeted deployment/firewall tests, including real Git fixtures where the branch exists only in a separate fork, environment selection, flag precedence, and invalid arguments rejected before SSH. Also ran the CLI against the actual `Kovbo/core` fork with SSH/install stubs: it fetched `f46b3892e577d89a7eb32f1e108422c44aa69c5c` and verified the heartbeat staging change before stopping at dependency installation. Bash syntax and ShellCheck pass. No additional Pod deployment is needed for this source-selection change; the archive/update path is unchanged from the live-tested PR #703.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ Cloned the repo on your computer? Build locally and deploy to an installed Pod:
 ```bash
 ./scripts/deploy POD_IP              # current checkout
 ./scripts/deploy POD_IP fix/my-fix   # branch from your clone's origin
+./scripts/deploy --repo Kovbo/core POD_IP fix/pod3-frozen-heartbeat  # fork branch
 ```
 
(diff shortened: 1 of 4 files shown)


ng:
@coderabbitai review

